PR #4, opened 2018-11-27
Ab 29 field validation

wojtek-krysiak:


admin/backend/adapters/abstract/instance.js
```diff
@@ -11,6 +11,7 @@ class AbstractInstance {
   constructor(params, model) {
     this.model = model
     this.params = params
+    this.errors = {}
```
mfka: Shouldn't it be defined outside of constructor ?  
const errors = {} ?

wojtek-krysiak: those are validation errors, and they are meant to be inside each Record/Document. when I put it in outside constructor it will be shared by all Records.

wojtek-krysiak: another thing is that I am thinking about removing totally AbstractInstance and put all `abstract logic` to AbstractModel

PR #12, opened 2018-11-29
STYLES ARCHITECTURE: functions, variables, mixins etc..

frontkickit:


admin/frontend/styles/abstracts/functions.scss
```diff
@@ -0,0 +1,3 @@
+@function rem($size-in-px, $context: $base-font-size) {
+  @return ($size-in-px / $context) * 1rem;
+}
```
michlask: should be sass not scss

PR #22, opened 2018-12-04
Ab 26

frontkickit:


admin/frontend/views/pages/list.pug
```diff
@@ -27,7 +27,12 @@ block content
             thead
               tr
                 for prop in currentResource.decorate().getListProperties()
-                  th.text-small= h._.startCase(prop.name())
+                  th(class= { 'main' : h.isMainColumn(prop.name())})
+                    - isSortedByCurrentPropName = sort.sortBy === prop.name()
+                    a(href=h.listUrl(currentResource, { sort: { sortBy: prop.name(), sortDirection: sort.direction === 'asc' && isSortedByCurrentPropName ? 'desc' : 'asc' }}))
```
michlask: [OPT] sort direction condition could be another variable to increase readability

admin/backend/adapters/mongoose/resource.js
```diff
@@ -23,8 +23,12 @@ class Resource extends BaseResource {
...
   }
 
-  async find(query, { limit = 20, offset = 0 }) {
-    const mongooseObjects = await this.MongooseModel.find({}).skip(offset).limit(limit)
+  async find(query, { limit = 20, offset = 0, sort }) {
+    const mongooseObjects = await this.MongooseModel
+      .find({})
+      .skip(offset)
+      .limit(limit)
+      .sort({[sort.sortBy]: sort.direction})
```
wojtek-krysiak: please add space after opening bracket: ({ [ (eslint)

admin/backend/controllers/resources-controller.js
```diff
@@ -71,10 +71,16 @@ class ResourcesController extends BaseController {
   async findRecords({ query }) {
     this.view.perPage = 10
     this.view.total = await this.view.currentResource.count()
-    this.view.page = query.page || 1
+    const { page, sortBy, sortDirection } = query
+    this.view.page = page || 1
+    this.view.sort = {
+      sortBy: sortBy || '_id',
```
wojtek-krysiak: in resource controller you cannot use raw field names like '_id' because in the future there will be other databases (which doesn't have _id). So here you have to iterate throught all model properties and find which property is an ID property (isId() method).

BUT
- i think that it should be filtered by default by first, visible in the list field :) - the simplest solution (this/view.currentResource.decorate().getListProperties() - first from that)

admin/backend/utils/view-helpers.js
```diff
@@ -4,18 +4,29 @@ const lodash = require('lodash')
 class ViewHelpers {
   constructor({ admin }) {
     this._admin = admin
-
     this.paginate = paginate
     this._ = lodash
     this.branding = this._admin.options.branding
   }
 
+  getObjectKeyWithValue(obj, key) {
```
wojtek-krysiak: not sure but maybe you can use url.parse(....): https://nodejs.org/api/url.html ?

admin/backend/controllers/resources-controller.js
frontkickit: I though that each document in a MongoDB collection has an _id - by default, but
{  $natural: 'asc' } query also returns the documents in the natural order

admin/backend/utils/view-helpers.js
frontkickit: I dont understand, but i think it works fine. We can have nested objects so recursion is needed

frontkickit: In this case I cant use url.parse

admin/backend/controllers/resources-controller.js
frontkickit: done

admin/backend/adapters/mongoose/resource.js
```diff
@@ -23,8 +23,14 @@ class Resource extends BaseResource {
     return this.MongooseModel.countDocuments()
   }
 
-  async find(query, { limit = 20, offset = 0 }) {
-    const mongooseObjects = await this.MongooseModel.find({}).skip(offset).limit(limit)
+  async find(query, { limit = 20, offset = 0, sort }) {
```
wojtek-krysiak: If I good remember test didn't work without sort having default `sort = {}`

PR #38, opened 2018-12-14
Ab 65 page builder

frontkickit:


src/backend/utils/page-builder.js
```diff
@@ -0,0 +1,264 @@
+/* eslint-disable class-methods-use-this */
```
wojtek-krysiak: General comment: I would move out all HTML strings somewhere else - the best would be to do this via PUG because we already have this in deps. You can even reuse Renderer class. 

For instance infoList would be:

```javascript
  addInfoList({ items, columns, offset, title, subtitle }) {
    // next line will go to partials/infoList.pug as a mixin, so its not even needed here
    // const itemsContent = items.map(item => this.getInfoListItem(item))
    const partialContent = new Renderer('partials/infoList', { items, columns, offset, title, subtitle }).render()
    this._pageContent.push(partialContent)
```

src/backend/utils/page-builder.js
```diff
@@ -0,0 +1,264 @@
...
+    const statusHtml = `
+      <div class="status ${status}">
+        ${status.toUpperCase()}
+      </div>`
+    return statusHtml
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  getInfoListItem(item) {
+    const { title, subtitle, content, status, date, imgSrc } = item
```
wojtek-krysiak: in general you can move destructuring parameters up to function definition and also add defaults there like: title = 'Example title'

src/frontend/views/pages/dashboard.pug
```diff
@@ -1,22 +1,33 @@
...
+      .overview
+        .columns
+          .column
+            .overview-title= customDashboard.title
+            .overview-subtitle= customDashboard.subtitle
+      != customDashboard.content
+      script.
+        const chartsModel = !{JSON.stringify(customDashboard.charts)}
+    else
+      section.is-fullheight.hero
```
wojtek-krysiak: the simplest way to make this more elegant would be: 

* add a new method to the PageBuilder: `#addDefaultDashboard` which will simply render this partial
* implement DefaultDashboard page which will extend PageBuilder, and us the above method in its build() implementation.
* use DefaultDashboard class in AdminBro defaults.

By having this change
* dashboard-controller.js wont have any ifs - there will always be _admin. DashboardPage class.
* you wouldn't have to have this if/else in pug template.
* and thats why this pug template could be used as generic template to render all the pages.

src/admin-bro.js
```diff
@@ -94,6 +96,7 @@ class AdminBro {
     const { databases, resources } = this.options
     const resourcesFactory = new ResourcesFactory(this, AdminBro.registeredAdapters)
     this.resources = resourcesFactory.buildResources({ databases, resources })
+    this.DashboardPage = options.dashboard || defaults.dashboard
```
wojtek-krysiak: defaults.dashboard is null so this line doesn't do anything

spec/backend/utils/page-builder.spec.js
```diff
@@ -0,0 +1,41 @@
...
+      expect(this.pageBuilder.convertedPageContent()).to.equal(null)
+    })
+
+    it('returns string of a HTML elements', function () {
+      this.pageBuilder.addBlock(options, this.types.succes)
+      expect(this.pageBuilder.convertedPageContent()).to.be.a('string')
+    })
+  })
+
+  // describe('#addBlock', function () {
```
wojtek-krysiak: if you don't need this - remove it.

src/backend/utils/page-builder.js
```diff
@@ -1,28 +1,142 @@
...
+ *     super(props)
+ *     this.title = 'My dashboard'
+ *   }
+ * 
+ *   async build() {
+ *     const articlesCount = await ArticleModel.countDocuments()
+ *     const publishedArticlesCount = await ArticleModel.find({ published: true }).countDocuments()
+ *     const unpublishedArticlesCount = articlesCount - publishedArticlesCount
+ *     const usersCount = await UserModel.countDocuments()
+ *     const comments = await CommentModel.find()
```
wojtek-krysiak: lets add `limit` set to like 10, to the query. I know this is an example but it hits the eye (you cannot fetch all models from db beacuse it will kill memory)

src/backend/utils/page-builder.js
```diff
@@ -1,28 +1,142 @@
...
+ *     })
+ *   }    
+ * }
  */
+
 class PageBuilder {
   /**
    *
    * @param  {AdminBro}     options.admin  current instance of AdminBro
+   * @param  {String[]}     _pageContent   array of html elements as String
```
wojtek-krysiak: those are params of the constructor. So only options.admin should be there. _pageContent for example is not a param - its an object property, so it should be documented like that.

Take a looka at documentation for resources property: https://github.com/SoftwareBrothers/admin-bro/blob/master/src/admin-bro.js#L86

src/backend/utils/page-builder.js
```diff
@@ -1,28 +1,142 @@
...
+   * @param  {String}       _pageHeader    html element as String
+   * @param  {String}       title          page title
+   * @param  {Object}       charts         mapped object contains chart's settings
+   * @param  {Object}       colorTypes     specific colors for html blocks
    */
   constructor({ admin }) {
     this._admin = admin
-
-    /**
-     * List of HTML elements as JavaScript string
```
wojtek-krysiak: this was fine

src/backend/utils/page-builder.js
```diff
@@ -47,8 +160,11 @@ class PageBuilder {
     }
   }
 
+  /**
+   * This method is responsible for building the page, should be overriden.
+   */
   build() {
-    throw new NotImplementedError()
+    throw new Error('You have to overwrite this method')
```
wojtek-krysiak: why not NotImplementedError?

src/backend/utils/page-builder.js
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  /** Adds info table element to the page content
+   * 
+   */
+  async addInfoTable({ title = '', columns = 12, items = [], offset = 0, headers = [] }) {
+    await this.addPartialContent('partials/infoTable', { title, columns, items, offset, headers })
+  }
+
+  /** Adds text box element to the page content
```
wojtek-krysiak: it should be:

```javascript
/** 
  * Adds text box element to the page content
  */
```

goes for all method descriptions

src/backend/utils/page-builder.js
```diff
@@ -0,0 +1,242 @@
...
+   * 
+   */
+  async addOverview(title = '', subtitle = '') {
+    this._pageHeader = await  new Renderer('partials/overview', { title, subtitle }).render()
+  }
+
+  /** Adds canvas chart to the page content
+   * 
+   */
+  async addChart(options) {
```
wojtek-krysiak: I would add info what could options be -> to be more precised: what could be options.config. (maybe link to chart js)

src/backend/utils/page-builder.js
```diff
@@ -0,0 +1,242 @@
...
+   */
+  async addDefaultDashboard() {
+    this._pageHeader =  await new Renderer('pages/defaultDashboard').render()
+  }
+
+  /** Adjusts overview content as a pageHeader
+   * 
+   */
+  async addOverview(title = '', subtitle = '') {
+    this._pageHeader = await  new Renderer('partials/overview', { title, subtitle }).render()
```
wojtek-krysiak: extra space between await and new

src/backend/utils/page-builder.js
```diff
@@ -0,0 +1,242 @@
...
+    if (this._pageContent.length > 0) {
+      return this._pageContent.join('')
+    }
+    return null
+  }
+
+  /** Adjusts default dashboard content as a pageHeader
+   * 
+   */
+  async addDefaultDashboard() {
```
wojtek-krysiak: method does not add - it sets value for _pageHeader so I would name the method set. But why we cannot reuse this.addPartialContent here?

src/backend/utils/page-builder.js
```diff
@@ -0,0 +1,242 @@
...
+    this._pageHeader =  await new Renderer('pages/defaultDashboard').render()
+  }
+
+  /** Adjusts overview content as a pageHeader
+   * 
+   */
+  async addOverview(title = '', subtitle = '') {
```
wojtek-krysiak: I would rename this to setOverview (it sets something not adds elements).

src/backend/utils/page-builder.js
```diff
@@ -0,0 +1,242 @@
...
+   */
+  build() {
+    throw new Error('You have to overwrite this method')
+  }
+
+  /** Returns string of html content elements
+   * @return {String}
+   */
+  convertedPageContent() {
+    if (this._pageContent.length > 0) {
```
wojtek-krysiak: maybe you can change lines 174-177 to simply `return this._pageContent.join('')` - in case of length === 0 -> you would simply return empty string

src/admin-bro.js
```diff
@@ -177,6 +181,8 @@ AdminBro.BaseRecord = BaseRecord
  */
 AdminBro.BaseProperty = BaseProperty
```
wojtek-krysiak: please add the description to this property (as we have for others) - it will generate links in documentation

src/backend/controllers/dashboard-controller.js
```diff
@@ -1,7 +1,11 @@
+/* eslint-disable no-unused-vars */
 const BaseController = require('./base-controller.js')
 
 class DashboardController extends BaseController {
-  index({params, query, payload}, response) {
+  async index({ params, query, payload }, response) {
+    // eslint-disable-next-line new-cap
+    const dashboard = this._admin.dashboardPage && new this._admin.dashboardPage(this._admin)
```
wojtek-krysiak: dashboardPage should be capitalized: this._admin.DashboardPage - its a class not an instance

src/backend/defaults/default-dashboard.js
```diff
@@ -0,0 +1,15 @@
+const PageBuilder = require('../utils/page-builder.js')
+const Renderer = require('../../backend/utils/renderer')
```
wojtek-krysiak: require('../utils/renderer') ?

src/backend/utils/page-builder.js
wojtek-krysiak: the same for ClientsModel

PR #43, opened 2018-12-20
AB-35 flatten fix

frontkickit:


src/backend/controllers/resources-controller.js
```diff
@@ -53,9 +54,18 @@ class ResourcesController extends BaseController {
         this.data.record,
       ))
     }
+    this.storePayloadData(payload)
     return this.render('pages/edit', this.data)
   }
 
+  storePayloadData(payloadData) {
+    Object.keys(payloadData).forEach(key => {
+      if(this.data.record.params[key]) {
```
wojtek-krysiak: accessing inner record members are not the responsibility of controllers. This should go to #update and #save methods within the BaseRecord

src/backend/adapters/base-record.js
```diff
@@ -22,10 +23,10 @@ class BaseRecord {
     this.resource = resource
 
     /**
-     * Actual record data stored as an object
+     * Actual record data stored as a flatten object
      * @type {Object}
      */
-    this.params = params
+    this.params = params && flatten(params)
```
wojtek-krysiak: I would add simple automated test

src/backend/adapters/base-record.js
```diff
@@ -92,6 +93,7 @@ class BaseRecord {
     } catch (e) {
       if (e instanceof ValidationError) {
         this.errors = e.errors
+        this.storePayloadData()
```
wojtek-krysiak: what it does? - asking because is called without the param

src/backend/adapters/base-record.js
```diff
@@ -63,6 +63,7 @@ class BaseRecord {
     } catch (e) {
       if (e instanceof ValidationError) {
         this.errors = e.errors
+        this.storePayloadData(params)
```
wojtek-krysiak: I would:
1. rename the method to something like: `storeParams` - describe what the method does from object perspective - only controller knows that this is from payload.
2. put this either in line 62 (before calling to update) or in finally statement. Updating the params is not a special case which happens only in case of error - it should occur always.

src/backend/adapters/base-record.js
frontkickit: its called without the params, because the save method doesn't contain parameters

spec/backend/adapters/base-record.spec.js
```diff
@@ -16,4 +16,44 @@ describe('Record', function () {
...
+
+    it('returns empty object if params are not passed to the constructor', function () {
+      const record = new Record()
+      expect(record.params).to.deep.equal({})
+    })
+
+    it('stores flatten object params', function () {
+      const record = new Record(this.params)
+      const expectedResult = { 'auth.login': 'login' }
+      expect(record.params).to.deep.equal(expectedResult)
```
wojtek-krysiak: I would write this in one line like that:

```javascript
expect(new Record({ auth: { login: 'login' } }).params).to.deep.equal({ 'auth.login': 'login' })
```

and you could remove entire before `beforeEach` along with lines 34 and 35 (10 lines in total less) - just saying :)

src/backend/adapters/base-record.js
```diff
@@ -63,6 +63,7 @@ class BaseRecord {
     } catch (e) {
       if (e instanceof ValidationError) {
         this.errors = e.errors
+        this.storeParams(params)
```
wojtek-krysiak: I would still put this above because we are not sure how update function will return params - I almost sure that in mongoose case they are not flattened. So after updating the record (without errors) you will have an unflatten params object.

PR #46, opened 2019-01-02
Ab 76 wysiwig

frontkickit:


src/frontend/views/mixins/form-group.pug
```diff
@@ -30,5 +30,9 @@ mixin formGroup(prop, value, error)
         label.label(for=prop.name)= prop.name()
         .control
           input.input(type='number' id=prop.name() name=prop.name() value=value)
+      when 'textEditor'
```
wojtek-krysiak: I would change that to "text" or "richtext" - to discuss

wojtek-krysiak:
I've updated a way of handling custom properties that it can be done for both sequelize and mongoose

PR #50, opened 2019-01-07
AB-80 Toggle in sidebar, mobile hamburger and manu animation

michlask:


src/frontend/scripts/app.js
```diff
@@ -1,12 +1,34 @@
...
+      const $target = $(event.target)
+      toggleMenu($target)
+      const status = $target.hasClass('icomoon-dropdown-open') ? 'open' : 'close'
+      localStorage.setItem($target.data('menu'), status)
+    });
+
+    $dropdownToggle.each((index, element) => {
+      const $element = $(element)
+      if (localStorage.getItem($element.data('menu')) === 'close') {
+        toggleMenu($element)
```
wojtek-krysiak: When I toggle all the menu items (that they are all hidden) - and refresh the page -> on load there is an animation of hiding - expected result: they should be hidden without any animation after refreshing the page.

PR #51, opened 2019-01-17
AB-104 options object

frontkickit:
current changes


wojtek-krysiak:
I did that in a separate PR

PR #53, opened 2019-02-04
Some small language corrections

moussaclarke:


moussaclarke:
Sorry - will try this again with correct indentations etc!

PR #87, opened 2019-05-16
Added styled-components theme provider

### LINKED ISSUE #80: Allow override of CSS theme values
Looking through the docs, it looks like we're very limited on customizations when it comes to brand, specifically only being able to override `logo` and `companyName`. Since all the branding colors are being pulled from `src/frontend/styles/variables.js` it stands to reason it should be simple to allow for an override for these default variables.

arronhunt:
Resolves #80 

The goal is to convert the current method of references theme variables to use styled-components' theme provider. This will allow for the default theme to be overridden and allow for more customization for admin-bro. 

**TODO:**
- [x] Use `<ThemeProvider>` and theme props for variables
- [x] Allow for a theme to be passed down via AdminBroOptions


arronhunt:
@wojtek-krysiak I also wanted to propose a few changes to the overall theme/design that I personally feel would be beneficial.

1. Changing default variable sizes to use the [8pt grid system](https://builttoadapt.io/intro-to-the-8-point-grid-system-d2573cde8632)
2. Convert to [rem units instead of pixel units](https://engageinteractive.co.uk/blog/em-vs-rem-vs-px) 
3. Changing some of the color values to improve contrast to meet accessibility standards (see image below)

Screen Shot 2019-05-15 at 9 38 46 PM

These are all changes that can easily be made to the `src/styles/variables.js` file and would greatly improve responsiveness and a11y of the application.

wojtek-krysiak:
@arronhunt next Friday at SoftwareBrothers we are hosting a hackaton and designers told me that they would like to update the design of AdminBro (on hackaton) :) - I already passed your comments to them so they will make an update.

wojtek-krysiak:
@arronhunt I've merged that to v0.9 branch. I will release it along with other bigger changes.

PR #97, opened 2019-05-29
Ab login improvements

michlask:


wojtek-krysiak:
@michlask I copied your solution, but I removed the background images

PR #179, opened 2019-10-18
Allowing redirect for non visible actions

ThiagoNP:
I created some components that needs to execute actions on the database from data inserted on a input (mainly approve/disapprove a account). 
These actions shouldn't be called directly, as the admin should review the data and insert a reason for disapproving.

This PR redirects the user if a custom action responds with a redirectUrl.


wojtek-krysiak:
I implemented this some time ago is another place - in action-button: https://github.com/SoftwareBrothers/admin-bro/blob/master/src/frontend/components/app/action-button.tsx#L100 So when user gives component: false in the given action and returns redirectUrl - then this code handles the redirect

PR #221, opened 2019-12-11
Adjust default theme colours to comply with WCAG 2.0 contrast guidelines

tomasztunik:
adjust main theme lightText, superLightBack and primary colours to comply with WCAG 2.0 contrast requirements at AA level

To keep the theme style and feel only lightness component on each colour was adjusted.

More details at https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html

This will help to make admin-bro more accessible to all the developers and users;
I've also noticed on my older laptop that current contrast on secondary text elements
was forcing me sometimes to tilt back the screen to be able to tell what's written.


tomasztunik:
Just noticed and read about https://github.com/SoftwareBrothers/admin-bro-dev should I move the PR there?

wojtek-krysiak:
thanks

PR #226, opened 2019-12-14
Improvement/branding options typings and ability to hide logo

tomasztunik:
**add ability to disable branding logo;** 
- it's possible to pass empty string but to make it more explicit you 
can also pass false as valid value

**improve typings for brandingOptions** 
- for AdmingBroOptionsWIthDefaults deriving them from BrandingOptions makes 
   it more explicit asto where properties are coming from and that we are not 
   extending the basic set

**improve overriding theme options by not forcing user to pass all options** 
- Introduced new utilityTypes colleciton witth PartialRecordWithValues
    for starters that allows creation partial records that do not allow
    including properties with nullish values, see the utilityTypes.ts file
- Implemented said type to theme options giving user the ability to override
    only selected subset of properties he is interested in

```json
      branding: {
        "companyName": "Company Name",
        "logo": false,
        "theme": {
          "colors": {
            "lightText": "#6c6d77",
            "lightBck": "#F8F8FA",
            "superLightBack": "#F1F1F5",
            "primary": "#718af4",
          },
        },
      },
```

before you had to copy all the properties, autocomplete is available for all possible values


tomasztunik:
@wojtek-krysiak Let me know if you are ok with adding this utilityTypes.ts file, wasn't sure where to put it.

Also thinking about the name for this type, would NonNullishPartialRecord be more explicit? with typescript 3.7 nullish everything will be the thing :)

src/frontend/utils/utilityTypes.ts
```diff
@@ -0,0 +1,13 @@
+
```
wojtek-krysiak: I would change the name to partial-record-with-values.type.ts (everywhere we have a dash notation). And please put it to src/utils

src/frontend/utils/utilityTypes.ts
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * PartialRecordWithValues
+ * Allows creating a type based on mapped type K that holds a value o type V (`string` by default)
+ * That makes all keys from mapped type optional but requires so that they have value.
+ * { } // ok
+ * { x: 1, y: 2 } // ok
+ * { x: 1 } // ok
+ * { x: 1, y: undefined } // error
```
wojtek-krysiak: please add `@private` jsdoc tag (so it won't be present in the generated documentation) and remove the extra line (11)

tomasztunik:
applied requested changes, will keep naming conventions and JSDoc usage in mind for the future!

Also decided to change the type name to `NonNullishPartialRecord` to make it more explicit as to what it does.

wojtek-krysiak:
Thanks!!! - I totally like it - deployed today along with admin-bro@1.5.0-beta.9

PR #305, opened 2020-02-19
fix: Removing of unnecessary height in new.tsx action

paweljar:


paweljar:
Unfortunately, when I'm setting `showInDrawer: false` for `edit` action I see:
image

PR #370, opened 2020-03-17
Fix/types and record validation

szymonpoltorakrst:


maurocen:
Does this fix the following issue?

```
Uncaught TypeError: Cannot read property 'constructor' of null
    at record-to-form-data.ts:25
```

szymonpoltorakrst:
@maurocen I think yes :)

wojtek-krysiak:
awesome - released in 2.1.0-beta.19

PR #389, opened 2020-04-02
Fix reference selects being sometimes empty

Co0sh:
Previously the `Edit` reference component only displayed the selected value when it was present in the `populated` field or was manually selected by the user. That is okay when editing fields or displaying them immediately after loading, assuming they're only one level deep. But AdminBro doesn't send `populated` fields for nested objects or objects in arrays. This means that for those properties the select displays no value even if it's actually correctly selected under the hood.

This PR converts `Edit` component to a function component and modifies it so that is uses the data loaded from the backend (which is already being used to fill the dropdown) to display currently selected option.

### DIFF
--- a/src/frontend/components/property-type/reference/edit.tsx
+++ b/src/frontend/components/property-type/reference/edit.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react'
+import React, { FC, useState, useEffect } from 'react'
 import Select from 'react-select/lib/Async'
 import { withTheme, DefaultTheme } from 'styled-components'
 
@@ -13,15 +13,10 @@ type SelectRecordEnhanced = SelectRecord & {
   record: RecordJSON;
 }
 
-class Edit extends React.Component<CombinedProps> {
-  constructor(props: CombinedProps) {
-    super(props)
-    this.loadOptions = this.loadOptions.bind(this)
-    this.handleChange = this.handleChange.bind(this)
-  }
+const Edit: FC<CombinedProps> = (props) => {
+  const { onChange, property, record, theme } = props
 
-  handleChange(selected: SelectRecordEnhanced): void {
-    const { onChange, property } = this.props
+  const handleChange = (selected: SelectRecordEnhanced): void => {
     if (selected) {
       onChange(property.name, selected.value, selected.record)
     } else {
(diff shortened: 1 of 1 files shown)


src/frontend/components/property-type/reference/edit.tsx
```diff
@@ -1,5 +1,7 @@
-import React, { ReactNode } from 'react'
-import Select from 'react-select/lib/Async'
+/* eslint-disable @typescript-eslint/no-use-before-define */
```
wojtek-krysiak, line 1: You can move loadOptions on the top and this won't be needed

src/frontend/components/property-type/reference/edit.tsx
```diff
@@ -13,67 +15,54 @@ type SelectRecordEnhanced = SelectRecord & {
...
       onChange(property.name, '')
     }
   }
 
-  async loadOptions(inputValue: string): Promise<Array<SelectRecordEnhanced>> {
-    const { property } = this.props
-    const api = new ApiClient()
+  const [options, setOptions] = useState<SelectRecordEnhanced[] | null>(null)
+  useEffect(() => {
+    loadOptions(property.reference!).then(setOptions)
```
wojtek-krysiak, line 30: options should be loaded also when user writes something in the input

src/frontend/components/property-type/reference/edit.tsx
```diff
@@ -1,5 +1,7 @@
-import React, { ReactNode } from 'react'
-import Select from 'react-select/lib/Async'
```
wojtek-krysiak, line 2: this is select with autocomplete

src/frontend/components/property-type/reference/edit.tsx
Co0sh: I know, but it's more readable when the main thing in the file - the component - is the first thing you see when you open it. I can follow this guideline, but it might be worth considering why we need it in the first place. Hoisting is a language feature for a reason.

src/frontend/components/property-type/reference/edit.tsx
Co0sh: Alright, so this requires a completely different approach. I'll revert these changes and start from scratch, then force push on this branch.

src/frontend/components/property-type/reference/edit.tsx
```diff
@@ -13,67 +13,78 @@ type SelectRecordEnhanced = SelectRecord & {
...
-      value: '',
-      label: '',
+  useEffect(() => {
+    if (!selectedValue && selectedId) {
+      setLoadingRecord(c => c + 1)
+      const api = new ApiClient()
+      api.recordAction({
+        actionName: 'show',
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        resourceId: property.reference!,
```
wojtek-krysiak, line 62: I would throw error on top of the function if property doesn't have reference. Now it can be also used by anyone with `BasePropertyComponent.Reference.Edit` so property.reference can be empty in this case.

src/frontend/components/property-type/reference/edit.tsx
```diff
@@ -13,67 +13,78 @@ type SelectRecordEnhanced = SelectRecord & {
...
   }
+  const error = record.errors && record.errors[property.name]
 
-  render(): ReactNode {
-    const { property, record, theme } = this.props
-    const error = record.errors && record.errors[property.name]
+  const selectedId = record?.params[property.name] as string | undefined
+  const [loadedRecord, setLoadedRecord] = useState<RecordJSON | undefined>()
+  const [loadingRecord, setLoadingRecord] = useState(0)
+  const selectedValue = record?.populated[property.name] ?? loadedRecord
```
wojtek-krysiak, line 45: question: loadedRecord shouldn't be first here? isn't it more important ten what is in record.populated?

src/frontend/components/property-type/reference/edit.tsx
```diff
@@ -13,67 +13,78 @@ type SelectRecordEnhanced = SelectRecord & {
...
 
     const records = await api.searchRecords({
       resourceId: property.reference as string,
       query: inputValue,
     })
-    return records.map((record: RecordJSON) => ({
-      value: record.id,
-      label: record.title,
-      record,
+    return records.map((r: RecordJSON) => ({
```
wojtek-krysiak, line 34: once I received a comment in CR (say hallo to @plorencrstit) regarding not using shortcuts like `r`, `m`, and I am trying not to do this anymore. So hmmm - could you please change it to `record` :)

plorencrstit: love it!

src/frontend/components/property-type/reference/edit.tsx
Co0sh: No, we only load that data because we don't always have it initially. `populated` field is only there if the property is top-level. Nested properties don't have `populated` so we need to load that data manually. But then when you select a value it will be added to `populated`, so no need to load manually data for it - it will be in `populated`. And the initially loaded data will be stale. That's why it's second, if it wouldn't then the component wouldn't change when selecting a value unless we made that extra request for data we already have.

src/frontend/components/property-type/reference/edit.tsx
Co0sh: That would only happen if AdminBro was misconfigured, right? I mean, it would need a user having access to a resource which references another resource that the user does not have access to, am I understanding this correctly?

src/frontend/components/property-type/reference/edit.tsx
Co0sh: `record` is already there though. @plorencrstit can I have a copy of that comment? I'd like to know your reasoning on that.

I'll change it to `optionRecord`.

plorencrstit: It is about readability.  Variables should tells us what are they. `r` says nothing. Maybe in this situation ot is obvious for you that you mean `record` but there could be many other situation when it is not. Then, it is just a goos practice to avoid variables as `e` instead of `error`. Not much to effort, but the effects could be great.

PR #397, opened 2020-04-03
Fix/upload files

wojtek-krysiak:


src/frontend/hooks/update-record.ts
```diff
@@ -5,9 +5,13 @@ import RecordJSON from '../../backend/decorators/record-json.interface'
 /**
  * Returns a function which takes a record and returns an updated record.
  *
- * @param property property that must be updated, supports nesting with dots
- * @param value value that must be set, undefined or null if deleting, will be flattened
- * @param refRecord if value is reference ID, this must be a record it's referencing to
+ * @param {string}      property    property that must be updated, supports nesting
```
wojtek-krysiak, line 8: added {} with types because otherwise JSDoc will throw errors.

src/frontend/hooks/update-record.ts
```diff
@@ -27,7 +31,7 @@ const updateRecord = (
   }
   // set new value
   if (value) {
-    if (typeof value === 'object') {
+    if (typeof value === 'object' && !(value instanceof File)) {
```
wojtek-krysiak, line 34: this is the actual fix - before function wanted to unflatten the File object (when file was given)

Issue #435, opened 2020-05-02
api-controller.js does not respect sort by column for associated tables


MichaelLeeHobbs:
**Describe the bug**
api-controller.js does not respect sort by column for associated tables

**Installed libraries and their versions**
```
"dependencies": {
    "admin-bro": "^2.2.12",
    "admin-bro-expressjs": "^2.0.5",
    "admin-bro-sequelizejs": "^0.4.1",
    "aws-sdk": "^2.666.0",
    "cron": "^1.7.2",
    "cron-parser": "^2.13.0",
    "dotenv": "^8.2.0",
    "express": "^4.17.1",
    "express-fileupload": "^1.1.7-alpha.3",
    "express-formidable": "^1.2.0",
    "fs-extra": "^8.1.0",
    "json2csv": "^4.5.4",
    "mirth-api": "^0.0.5",
    "moment": "^2.24.0",
    "morgan": "^1.10.0",
    "node-fetch": "^2.6.0",
    "nodemailer": "^6.3.1",
    "pg": "^7.14.0",
    "sequelize": "^5.21.2",
    "swagger-jsdoc": "^4.0.0",
    "vm2": "^3.8.4"
  }
```

**To Reproduce**

**Expected behavior**
Respect sortBy option as failing to do this causes issues with virtual columns

**AdminBroOptions with schema**
```
{
        databases: [db],
        resources: [
            {
                resource: db.Report,
                options: {
                    properties: {
                        script: {
                            type: 'textarea',
                            isVisible: { list: false, filter: false, show: true, edit: true },
                        },
                    }
                },
            },
            {
                resource: db.Database,
                options: {
                    properties: {
                        alias: {isTitle: true},
                    }
                },
            },
            {
                resource: db.Provider,
                options: {
                    sort: {sortBy: 'familyName'},
                    editProperties: ['nationalProviderIdentifier', 'familyName', 'givenName', 'additionalName'],
                    listProperties: ['nationalProviderIdentifier', 'name', 'createdAt', 'updatedAt'],
                    properties: {
                        name: {isDisabled: true},
                    }
                },
            },
        ],
        rootPath: '/admin',
    }
```

**Desktop (please complete the following information if relevant):**
 - OS: Docker FROM node:12.8.0-alpine
 - Browser chrome

**Additional context**
see: https://github.com/SoftwareBrothers/admin-bro-sequelizejs/issues/26

** FIX **
api-controller.js
```
[...]

PR #442, opened 2020-05-06
fixes linter issue for #435

MichaelLeeHobbs:


MichaelLeeHobbs:
Will refork latter and fix linter issue

wojtek-krysiak:
ok - I spotted one other issue so I will do this btw and release them both

PR #487, opened 2020-06-19
Allow overriding resource list in sidebar component

JonaszJestem:


src/frontend/components/app/sidebar/sidebar-resource-section.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FC } from 'react'
+import groupResources from './utils/group-resources'
+import SidebarParent from './sidebar-parent'
+import ResourceJSON from '../../../../backend/decorators/resource-json.interface'
+import allowOverride from '../../../hoc/allow-override'
+
+type SidebarResourceSectionProps = {
```
wojtek-krysiak, line 7: we will also need to export this type from AdminBro

src/frontend/components/app/sidebar/sidebar-resource-section.tsx
```diff
@@ -0,0 +1,26 @@
...
+    resources: ResourceJSON[];
+}
+
+const SidebarResourceSection: FC<SidebarResourceSectionProps> = ({ resources }) => {
```
wojtek-krysiak, line 11: please add JSDoc tag here so it will be seen in docs

Issue #515, opened 2020-07-11
remove reference properties value


hvasc:
Hi, I want adding reference type component's remove button (assign null attribute for reference value)

can I customize admin-bro's default component?
otherwise can import reference type's edit component in my formgroup?

leibowitz:
You can customise the [`edit reference`](https://github.com/SoftwareBrothers/admin-bro/blob/26c637829c5576bb7a221294343f6a78d8da2831/src/frontend/components/property-type/reference/edit.tsx) type

But I believe this would be best handled by passing [`isClearable`](https://react-select.com/props) to the [react-select](https://github.com/JedWatson/react-select) [`Select`](https://github.com/SoftwareBrothers/admin-bro/blob/26c637829c5576bb7a221294343f6a78d8da2831/src/frontend/components/property-type/reference/edit.tsx#L79-L88) component

hvasc:
> You can customise the [`edit reference`](https://github.com/SoftwareBrothers/admin-bro/blob/26c637829c5576bb7a221294343f6a78d8da2831/src/frontend/components/property-type/reference/edit.tsx) type
> 
> But I believe this would be best handled by passing [`isClearable`](https://react-select.com/props) to the [react-select](https://github.com/JedWatson/react-select) [`Select`](https://github.com/SoftwareBrothers/admin-bro/blob/26c637829c5576bb7a221294343f6a78d8da2831/src/frontend/components/property-type/reference/edit.tsx#L79-L88) component when a field is nullable

thank you very much!

leibowitz:
I meant this should be handled in the existing edit component, in my opinion. 

You can try sending a PR with a code change that checks the property (if nullable) and pass `isClearable` as props to the `Select` component in that case.

Alternatively you could just duplicate the whole component and use a version with `isClearable` added to the props. But that sounds like overkill

hvasc:
> I meant this should be handled in the existing edit component, in my opinion.
> 
> You can try sending a PR with a code change that checks the property (if nullable) and pass `isClearable` as props to the `Select` component in that case.
> 
> Alternatively you could just duplicate the whole component and use a version with `isClearable` added to the props. But that sounds like overkill

thank very much. I'll try it

wojtek-krysiak:
it duplicates #161

PR #520, opened 2020-07-20
reference type clear selected value

hvasc:
#515 
this version's reference type is just search and select only.
so add [`isClearable`](https://react-select.com/props) to the [react-select](https://github.com/JedWatson/react-select) [`Select`](https://github.com/SoftwareBrothers/admin-bro/blob/26c637829c5576bb7a221294343f6a78d8da2831/src/frontend/components/property-type/reference/edit.tsx#L79-L88) in reference's edit component.
then developer want setting up null using [this solution](https://github.com/SoftwareBrothers/admin-bro/issues/342#issuecomment-621442262)

image
image


wojtek-krysiak:
Thanks for this PR @hvasc. I was looking at it but I think that it should be a more systematic approach for removing reference which will clear the reference also in the database when the user clears the select.
It will be quite complicated to achieve because we don't know which elements are cleared and which simply doesn't appear in the UI (isVisible set to false) - that is why this task is laying there for a while :)

PR #527, opened 2020-07-25
Beta

wojtek-krysiak:


leibowitz:
@wojtek-krysiak thanks for merging, cleaning and releasing. Though I noticed you removed [this line](https://github.com/leibowitz/admin-bro/blob/af41cbe6e0c9892c758b58a5264df2d18338b3a9/src/frontend/components/design-system/molecules/date-picker.tsx#L139) probably while resolving conflicts

As you can see, the [release](https://github.com/SoftwareBrothers/admin-bro/blob/v2.9.0/src/frontend/components/design-system/molecules/date-picker.tsx#L135-L139) doesn't include that line, which means passing custom props doesn't work in 2.9.0

wojtek-krysiak:
yes - you are correct. I will try to fix in 2.9.1

wojtek-krysiak:
but wait. I added `CustomProps & ...` (https://github.com/SoftwareBrothers/admin-bro/blob/v2.9.0/src/frontend/components/design-system/molecules/date-picker.tsx#L122).

it should allow you to pass any props allowed by datepicker since it CustomProps is `Partial<Omit<ReactDatePickerProps, 'value' | 'disabled' | 'onChange'>>`

Does it work?

leibowitz:
Don't think, since I tried with 2.9.0 (beta and release) and that's what lead me to look at the code and comment on here.

leibowitz:
Did I use a specific version of typescript? I'm not familiar with how TS works, but from what I can tell in [package.json](https://github.com/SoftwareBrothers/admin-bro/blob/fc830deeda5a661792fe2ca16f1f98b5a8e51cd0/package.json#L130) admin-bro uses [@types/react-datepicker](https://www.npmjs.com/package/@types/react-datepicker) v3.1.0. I installed it locally and check the types, and I can see all these options. But maybe you were using a different version?

PR #533, opened 2020-07-30
fix: load reference object inside onChange callback when value updated

leibowitz:
I was attempting to change the reference property value on a record using `onChange` (passed from props) on an edit form. For some reason the field value is updating, but not the selected value in the react-select component.

The weird part is, it was updating correctly after calling `onChange` once. But subsequent calls seems to fail to update the react-select selected value. I found this is due to the check on `selectedValue`, which is being set after the first call.

Adding a comparison on record.id seems to fix the issue.


leibowitz:
I think it's actually working without this change, just by setting the related record in the `record.populated[field]`. Is that correct?

wojtek-krysiak:
@Co0sh I think you had a similar issue?

PR #537, opened 2020-08-07
Improve performance of text inputs

Co0sh:
This makes inputs only commit changes when they loose focus, instead of on every key press. This eliminates the janky feeling when typing in huge records while on a relatively slow device.


Co0sh:
@wojtek-krysiak should I target this PR at `master` or some other branch?

PR #554, opened 2020-08-24
feat: add step, min and max props for number properties

maurocen:
# What this PR does
- Adds `step`, `min` and `max` props when a property is of type `number` or `float`.
  - `step` defaults to `1`
  - `min`/`max` default to `Number.MIN_SAFE_INTEGER`/`Number.MAX_SAFE_INTEGER` or `Number.MIN_VALUE`/`Number.MAX_VALUE`  depending on the type being `number` or `float`
- Properties are pulled from `custom` object in property declaration.

### DIFF
--- a/src/frontend/components/property-type/default-type/edit.tsx
+++ b/src/frontend/components/property-type/default-type/edit.tsx
@@ -10,6 +10,26 @@ import usePrevious from '../../../utils/usePrevious'
 
 type CombinedProps = EditPropertyProps & {theme: DefaultTheme}
 
+const getNumberProps = (type, options) : object => {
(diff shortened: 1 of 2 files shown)


wojtek-krysiak:
It looks ok, but unfortunately, I cannot merge it :(. We have a plan for handling integers in a full approach:

* create a new `number` type for edit and filter
* create dedicated react component for numbers
* update filter logic so it can handle `greater than`, `less than`, `less or equal than` etc

I will leave this open so it will be an inspiration for this feature.

aleccool213:
Has this been done? I dont see why Admin Bro saves numbers to a database as a string.

Issue #352, opened 2020-03-06
AdminBro crashes when saving record with an empty array


Co0sh:
**Steps to reproduce:**

1. Install AdminBro with Express and Mongoose adapters
2. Have resource with an array of objects (let's call them `Foo`) as one of the properties. Those nested objects have to contain an array of objects (let's call them `Bar`) as one of their properties.
3. Click "Create new" button (`new` action).
4. Fill the form with some data, including one `Foo` object in the array.
5. Fill the `Foo` object with some data, including one `Bar` object in the array.
6. Remove all `Bar` objects from the array. _(This is what triggers the bug. If the `Bar` array was empty and the user wouldn't touch it, the bug would not appear. Adding and removing items triggers the bug.)_
7. Save the record. AdminBro crashes the whole application. Not just the request, the whole app goes down.

**Stack trace:**

```
/home/user/Projects/some-project/node_modules/mongoose/lib/error/objectParameter.js:25
    Error.captureStackTrace(this);
          ^
ObjectParameterError: Parameter "obj" to Document() must be an object, got 
    at new ObjectParameterError (/home/user/Projects/some-project/node_modules/mongoose/lib/error/objectParameter.js:25:11)
    at EmbeddedDocument.Document (/home/user/Projects/some-project/node_modules/mongoose/lib/document.js:87:11)
    at EmbeddedDocument [as constructor] (/home/user/Projects/some-project/node_modules/mongoose/lib/types/embedded.js:42:12)
    at new EmbeddedDocument (/home/user/Projects/some-project/node_modules/mongoose/lib/schema/documentarray.js:116:17)
    at cb (/home/user/Projects/some-project/node_modules/mongoose/lib/schema/documentarray.js:252:26)
    at DocumentArrayPath.SchemaType.doValidate (/home/user/Projects/some-project/node_modules/mongoose/lib/schematype.js:1046:12)
    at DocumentArrayPath.doValidate (/home/user/Projects/some-project/node_modules/mongoose/lib/schema/documentarray.js:201:37)
    at /home/user/Projects/some-project/node_modules/mongoose/lib/document.js:2323:9
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

**Example object `Foo`:**

```ts
interface Foo {
  name: string;
  nested: Bar;
}

interface Foo {
  title: string;
  bugs: Baz; // this is the part triggering the bug
}

interface Baz {
  anything: number;
}
```

**More info:**

[...]

Issue #304, opened 2020-02-19
Edit resources with NULL values


telmobarros:
I can't edit resources with NULL values that I desire to keep as NULL.

To begin with I receive this Warning in the browser. (From React I suppose)

> Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.

Then if I try to save after changing only one value it just clears the form. If I want to update I have to fill all the input fields, even the ones I desire to keep NULL.

wojtek-krysiak:
can you paste a schema and please write which adapter you use (mongoose/typeorm/sequelize)?

telmobarros:
This model for instance, using sequelize adapter. I am able to create a new Partner leaving website as an empty string and its converted to NULL after inserting in the database. But when I try to edit I am not able to do it at all. I tried forcing the allowNull property to true but I get the same error

```
'use strict';
module.exports = (sequelize, DataTypes) => {
	const Partner = sequelize.define('Partner', {
		name: {
			allowNull: false,
			type: DataTypes.STRING(50)
		},
		logo: {
			allowNull: false,
			type: DataTypes.STRING
		},
		website: {
			type: DataTypes.STRING
		}
	}, {});
	return Partner;
};
```

mjlawrence83:
I've seen the #370 fix go in, but it changes all NULLs to be blank string.

I'm planning to have admin-bro working with an existing MSSQL DB via sequelize, and at the moment it will be changing pre-existing records replacing NULLs with ''. I can't afford to allow data changing unexpectedly. Is there a plan to have it handle NULLs or will it continue as its currently implemented? (I can't make any changes via the models I'm using sequelize-auto to generate them on the fly)

wojtek-krysiak:
@mjlawrence83 wow - this is an actual very good solution. We will have to implement this in the core.

MichaelLeeHobbs:
Just sharing the approach I took to resolve this issue.

```
/* jshint indent: 1 */

const NAME = 'Email'

module.exports = (sequelize, DataTypes) => {
    let schema = {
        templateData: {
            type: DataTypes.JSON,
            allowNull: true,
        },
        from: {
            type: DataTypes.STRING,
            allowNull: false,
            unique: false,
            validate: {
                isEmail: true
            }
        },
        sender: {
            type: DataTypes.STRING,
            allowNull: true,
            unique: false,
            validate: {
                isEmail: true
            }
        },
        replyTo: {
[...]

Issue #161, opened 2019-10-08
Allow to clear <select> field


wojtek-krysiak:
When there is a select field (i.e. reference) there should be an option to clear it's value (when field is not required)

In this task it would be perfect if we can rewrite the select component to be in our design system

PR #560, opened 2020-08-26
Beta

wojtek-krysiak:
fixes: #352, #304 and #161

What I did I:
* moved all the frontend code related to handling record state to `useRecord` folder - along with useRecord hook
* removed obsolete hooks, which were replaced by useRecord
* handled passing empty arrays, objects and nulls by changing them to fixed constants on the frontend and bringing them back to original values on the backend


src/backend/utils/request-parser.spec.ts
```diff
@@ -19,12 +19,6 @@ describe('RequestParser', function () {
         return newProperty
       },
     } as BaseResource
-
-    it('converts empty string to an empty array', function () {
```
wojtek-krysiak, line 23: requst parser changes now constant to empty array.

src/backend/utils/request-parser.ts
```diff
@@ -14,27 +15,29 @@ import BaseResource from '../adapters/base-resource'
...
-        if (value === '') { return { ...memo, [path]: null } }
+        if (value === '' || value === null) { return { ...memo, [path]: null } }
       }
       if (property.type() === 'string') {
         const availableValues = property.availableValues()
         if (availableValues && !availableValues.includes(value) && value === '') {
           return { ...memo, [path]: null }
         }
       }
-      if (property.isArray() && value === '') {
```
wojtek-krysiak, line 35: line 24 handles that

src/frontend/components/property-type/reference/edit.tsx
```diff
@@ -24,7 +24,7 @@ const Edit: FC<CombinedProps> = (props) => {
     if (selected) {
       onChange(property.name, selected.value, selected.record)
     } else {
-      onChange(property.name, '')
+      onChange(property.name, null)
```
wojtek-krysiak, line 27: now null can be handled by onChange so we can clear select fields

src/frontend/hooks/use-record/record-to-form-data.ts
```diff
@@ -0,0 +1,40 @@
...
+ * FormData is required because we are sending files via the wire. But it has limitations.
+ * Namely it can only transport files and strings. That is why we have to convert some
+ * standard types like NULL to constants so they can be property converted back by the backend.
+ * And thus properly handled.
+ *
+ *
+ * @param   {RecordJSON}  record
+ * @return  {FormData}
+ */
+export default function recordToFormData(record: RecordJSON): FormData {
```
wojtek-krysiak, line 19: this function allows onChange to transport `string`, `[]`, `null` and `{}` to the backend

src/frontend/hooks/use-record/update-record.ts
```diff
@@ -0,0 +1,78 @@
+import flat from 'flat'
```
wojtek-krysiak, line 1: left this file as it was but moved to `use-record/`

src/frontend/hooks/use-record/update-record.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  // set new value
+  if (typeof value !== 'undefined') {
+    if (typeof value === 'object' && !(value instanceof File) && value !== null) {
+      const flattened = flat.flatten(value) as any
+      if (Object.keys(flattened).length) {
+        Object.keys(flattened).forEach((key) => {
+          paramsCopy[`${property}.${key}`] = flattened[key]
+        })
+      } else if (Array.isArray(value)) {
```
wojtek-krysiak, line 56: I added that, so before when user gave `{}` or `[]` nothing happened. Now it is stored as "[]" or "{}" and then in toFormData converted to constants so it can be passed in the correct form to the backend

src/frontend/hooks/use-record/update-record.ts
```diff
@@ -0,0 +1,78 @@
...
+        paramsCopy[property] = []
+      } else {
+        paramsCopy[property] = {}
+      }
+    } else {
+      paramsCopy[property] = value
+    }
+  }
+
+  if (selectedRecord) {
```
wojtek-krysiak, line 66: before this was in the above if. But I think that users may set populated record even when value is null like this:

onChange('populatedPropertyName', null, record).

to reset the param but to maintain the value in populated. This is not super likely but possible, and I don't see a place when it could break something

src/backend/utils/request-parser.ts
```diff
@@ -14,27 +15,29 @@ import BaseResource from '../adapters/base-resource'
 const RequestParser = (originalRequest: ActionRequest, resource: BaseResource): ActionRequest => {
   const { payload: originalPayload } = originalRequest
 
-  const payload = Object.entries(originalPayload || {}).reduce((memo, [path, value]) => {
+  const payload = Object.entries(originalPayload || {}).reduce((memo, [path, formValue]) => {
     const property = resource.property(path)
```
wojtek-krysiak, line 19: here is a known issue - `resource.property` doesn't take into consideration nested properties and options. So when user changes type of the property it wont be affected here. The same goes for nested properties - they wont be properly converted back.

I am on it

PR #564, opened 2020-08-27
fix: onChange can also update other fields

wojtek-krysiak:
@Co0sh there was an error where users couldn't update properties with onChange.

Baicaly this component didn't work:

```
import React from 'react'
import { EditPropertyProps } from 'admin-bro'
import { Button, Box } from '@admin-bro/design-system'

const ValueTrigger: React.FC<EditPropertyProps> = (props) => {
  const { onChange, record } = props

  const handleClick = (): void => {
    onChange({
      ...record,
      params: {
        ...record.params,
        name: 'my new name',
      },
    })
  }

  return (
    <Box mb="xxl">
      <Button data-testid="name-button" type="button" onClick={handleClick}>Set Name</Button>
    </Box>
  )
}

export default ValueTrigger
```

Name property is a string and inside there is an inner state which was only updated from the external `params` on component mount.

I've fixed that and will merge it, but please verify if this is ok.


src/frontend/components/property-type/default-type/edit.tsx
```diff
@@ -52,13 +51,11 @@ const TextEdit: FC<CombinedProps> = (props) => {
   const propValue = record.params?.[property.name] ?? ''
   const [value, setValue] = useState(propValue)
 
-  const previous = usePrevious(propValue)
```
wojtek-krysiak, line 55: removed usePrevious because the previous state is in value

src/frontend/components/property-type/default-type/edit.tsx
```diff
@@ -52,13 +51,11 @@ const TextEdit: FC<CombinedProps> = (props) => {
...
   useEffect(() => {
-    // this means props updated
-    if (propValue !== previous) {
+    if (value !== propValue) {
       setValue(propValue)
     }
-  }, [])
+  }, [propValue])
```
wojtek-krysiak, line 58: added that which invalidates when external param was updated

Issue #601, opened 2020-09-14
Mobile Sidebar keep opened when navigating to another link


johnyvelho:
**Describe the bug**
When in mobile, once clicking to navigate to another resource, the sidebar remains opened, which is hard to the user know that the page it was loaded already. It looks like nothing happened.

**To Reproduce**
1. Open the admin in mobile mode
2. Open the sidebar
3. Click to navigate to another page
4. Siderbar remains open

**Expected behavior**
To be clear to the user that the page was loaded, after clicking to navigate to another page, we should close the sidebar.

_I will open a PR for it._

PR #602, opened 2020-09-15
fix: [#601] Sidebar state is now stored in Redux Store

johnyvelho:
Fix for #601.

Sidebar state is now controlled on Redux State.
Created a React Hook for Sidebar that can be called anywhere


wojtek-krysiak:
thanks - this looks really good

PR #635, opened 2020-10-09
fix: fix BaseRecord.param(path) returning [null]

dziraf:
This is a fix to: https://github.com/SoftwareBrothers/admin-bro/issues/634
However, I've no idea why `flat.get(...)` would return `[null]`.

### DIFF
--- a/src/backend/adapters/record/base-record.ts
+++ b/src/backend/adapters/record/base-record.ts
@@ -53,7 +53,11 @@ class BaseRecord {
    * @deprecated in favour of {@link BaseRecord#get} and {@link BaseRecord#set} methods
    */
   param(path: string): any {
-    return flat.get(this.params, path)
+    const paramValue = flat.get(this.params, path)
+    if (Array.isArray(paramValue) && paramValue.every(v => !v)) {
+      return null
+    }
+    return paramValue
   }
 
   /**


src/backend/adapters/record/base-record.ts
```diff
@@ -53,7 +53,11 @@ class BaseRecord {
    * @deprecated in favour of {@link BaseRecord#get} and {@link BaseRecord#set} methods
    */
   param(path: string): any {
-    return flat.get(this.params, path)
+    const paramValue = flat.get(this.params, path)
+    if (Array.isArray(paramValue) && paramValue.every(v => !v)) {
```
dziraf, line 57: Unsure about `every` vs `some`. I don't know if it's possible to get more than one `null` in this issue

wojtek-krysiak:
@dziraf thanks for the update. I would make 2 changes:

1. move this to the flat.get since it should replace all the places where we flat/unflat properties
2. write unit test in get: https://github.com/SoftwareBrothers/admin-bro/blob/bb9de777d67e3425928a2b6a8825626c7f9bc4bd/src/utils/flat/get.spec.ts#L9-L9 covering this particular issue

wojtek-krysiak (changes requested):
@dziraf thanks for the update. I would make 2 changes:

1. move this to the flat.get since it should replace all the places where we flat/unflat properties
2. write unit test in get: https://github.com/SoftwareBrothers/admin-bro/blob/bb9de777d67e3425928a2b6a8825626c7f9bc4bd/src/utils/flat/get.spec.ts#L9-L9 covering this particular issue

wojtek-krysiak:
@dziraf but wait - I think I fixed that 2 days ago here: https://github.com/SoftwareBrothers/admin-bro/commit/bb9de777d67e3425928a2b6a8825626c7f9bc4bd#diff-15b392d592d49fff9fe23078e53a9474R22

so if the value was null or 0 (before) it didn't return it directly but instead, it went further in the code eventually resolving to Array of nulls

wojtek-krysiak:
closing because it was already fixed

PR #647, opened 2020-10-23
fix: array manipulaton in deeply nested properties

wojtek-krysiak:
This is the first part of the update. The second will be the population of nested arrays.

What was done:
* all 'old' `PropertyJSON` fields send from the backend now becomes `BasePropertyJSON`. 
* `BasePropertyJSON` is `Omit<PropertyJSON, 'path'>`
* both of them got new property: `propertyPath` (old path) which is the "class" path.

So we have 3 fields defining the property path: 
* `propertyPath` - "class" path which can be used in PropertyOptions: `path.to.my.property`
* `path` - real instance path, which can be used to access actual record: `path.to.my.0.property` (they differs when it comes to arrays)
* `name` - the last part of the path, used to compute one from the other.

So data which goes from the bacend doesn't have `path` but it is require in custom components. So... All the real path calculation is done in BasePropertyCompoent.

Firstly it assume that what it receives on the root level: `path` in `path.to.my.0.property` example is also the path. And then it starts iterating deeper. Both Array and Mixed property types extends this path when they get the property and it is done recursively.

The last problem was with removing items from the array because when we remove one element we have to adjust index of all other elements which was already there. And... adjust the index of the populated property. This was done by adding new flat helper: removePath

### DIFF
--- a/example-app/src/employees/employee.entity.js
+++ b/example-app/src/employees/employee.entity.js
@@ -1,5 +1,33 @@
 const mongoose = require('mongoose')
 
+const Skill = new mongoose.Schema({
+  name: String,
+  level: {
+    type: String,
+    enum: ['junior', 'middle', 'senior'],
+  },
+  Profession: {
+    type: mongoose.SchemaTypes.ObjectId,
+    ref: 'Profession',
+  },
+})
+
+const OtherSchema = new mongoose.Schema({
+  name: String,
+  arrayed: {
+    type: [Skill],
+  },
+})
+
+const Skills = new mongoose.Schema({
+  softShills: [{
+    type: Skill,
+  }],
+  hardSkills: [{
+    type: Skill,
+  }],
+})
+
 const EmployeeSchema = new mongoose.Schema({
   name: {
     type: String,
@@ -16,10 +44,21 @@ const EmployeeSchema = new mongoose.Schema({
     type: mongoose.SchemaTypes.ObjectId,
     ref: 'Company',
   },
+  interests: {
+    type: [String],
+  },
   professions: [{
     type: mongoose.SchemaTypes.ObjectId,
     ref: 'Profession',
   }],
+
+  Skills: {
+    type: Skills,
+  },
+
+  otherField: {
+    type: [OtherSchema],
+  },
 })
 
 const Employee = mongoose.model('Employee', EmployeeSchema)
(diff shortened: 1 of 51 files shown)


example-app/src/employees/employee.entity.js
```diff
@@ -1,5 +1,33 @@
 const mongoose = require('mongoose')
 
+const Skill = new mongoose.Schema({
```
wojtek-krysiak, line 3: just a crazy example in example app which allows me to validate everything live

src/backend/decorators/property/property-decorator.ts
```diff
@@ -23,7 +23,7 @@ class PropertyDecorator {
    * This path serves as a key in {@link PropertyOptions} to identify which
    * property has to be updated
    */
-  public path: string
+  public propertyPath: string
```
wojtek-krysiak, line 26: to not to confuse i renamed path to propertyPath in the property decorattor

src/frontend/components/property-type/array/edit.spec.tsx
```diff
@@ -31,6 +31,7 @@ describe('<PropertyType.Array.Edit />', function () {
         ItemComponent={ItemComponent as unknown as typeof React.Component}
         onChange={onChange}
         testId="some-test-id"
+        filter={{}}
```
wojtek-krysiak, line 34: type error

src/frontend/components/property-type/array/edit.tsx
```diff
@@ -64,25 +53,26 @@ const InputsInSection: React.FC<EditProps> = (props) => {
     return false
   }, [record, onChange, property])
 
-  const removeItem = useCallback((i, event: MouseEvent): boolean => {
-    const newItems = [...items]
-    newItems.splice(i, 1)
-    onChange(property.path, newItems)
+  const removeItem = useCallback((event: MouseEvent, subProperty: PropertyJSON): boolean => {
+    const newRecord = removeSubProperty(record, subProperty.path)
```
wojtek-krysiak, line 57: added this function: removeSubProperty

src/utils/flat/remove-path.doc.md
```diff
@@ -0,0 +1,56 @@
+This method removes the given path from the flatten object.
```
wojtek-krysiak, line 1: this and the next functons are the one who does the magic

ayushbhat:
Great Job!!!

Issue #650, opened 2020-10-23
Add the ability to use external react dependecies.


LahmerIlyas:
Hi, thx for this awesome project, it saved a lot of time for me. I'm just wondering why is it hard to use external react dependencies, I tried to use many external packages, but with no success.

wojtek-krysiak:
we use an old version of rollup bundler - there are tasks in backlog to change that but not scheduled right now #614 #359 #185 #575

wojtek-krysiak:
I added more info to https://github.com/SoftwareBrothers/admin-bro/issues/575 - let it be the core task fro bundler long term improvements. Closing this one and lets move communication there

Issue #638, opened 2020-10-12
Richtext store in markdown


ArcaneDiver:
**Problem**
I sould save the text from a `richtext` property in the markdown format because the Quill's html it is pratically end at itself

If you know any ways to solve this please tell them to me.

**Proposal**
You migth migrate to a more modern rich text editor like `SlateJs` which easily supports Markdown serialize

wojtek-krysiak:
hmm - interesting. We implemented quill 2 years ago and haven't thought about changing it to something else. But this slatejs looks promising. Do you have more experience with it?

Until then you can try to use it on your own - simply create your own component for given property.

ArcaneDiver:
I have no experience on it.
May you reference where i could look for implement my own component?

wojtek-krysiak:
this is a big thing for a PR to the core right away (you will have to PR this to @admin-bro/design-system). But you can make it a feature as @admin-bro/upload or @admin-bro/password. I can then publish it on our documentation page.

ArcaneDiver:
I tried to implement by my own and i recived errors on bundle similiar to this issue #492. Then i think that in the next days i'll investigate and make a PR to resolve those kind of problems.

wojtek-krysiak:
I think I fixed that in latest beta so just upgrade and you should be able to import slate without a problem. https://github.com/SoftwareBrothers/admin-bro/releases/tag/v3.3.0-beta.25

`yarn add admin-bro@v3.3.0-beta.25`

ArcaneDiver:
Awesome then i'll try It as soon as possible

wojtek-krysiak:
So now this can be resolved by a custom AdminBro Feature. that is why I am closing this one, and opening this: #654

PranaytheSingh:
,mnn

Issue #492, opened 2020-06-26
Can't include external dependencies in React components


BeaRRRRR:
I trying to create a custom dashboard component in which I'll embed Atlas Charts using [this npm package](https://www.npmjs.com/package/@mongodb-js/charts-embed-dom).

Creating a custom dashboard works fine by itself, but when I try to require this dependency:

```
import React, { useEffect } from 'react';
import ChartsEmbedSDK from '@mongodb-js/charts-embed-dom';

const Dashboard = (props) => {
    useEffect(() => {
        const sdk = new ChartsEmbedSDK({
            baseUrl: 'https://charts.mongodb.com/charts-charts-fixture-tenant-zdvkh',
        });
        const chart = sdk.createChart({
            chartId: '48043c78-f1d9-42ab-a2e1-f2d3c088f864',
        });
    },[])

    return (
        <div>
        </div>
    )
}

export default Dashboard;
```

It shows a bunch of errors in the console

```
'buffer' is imported by node_modules/@mongodb-js/charts-embed-dom/node_modules/bson/dist/bson.esm.js, but could not be
resolved – treating it as an external dependency
'tty' is imported by node_modules/debug/src/node.js, but could not be resolved – treating it as an external dependency
'util' is imported by node_modules/debug/src/node.js, but could not be resolved – treating it as an external dependency
'fs' is imported by node_modules/debug/src/node.js, but could not be resolved – treating it as an external dependency
'net' is imported by node_modules/debug/src/node.js, but could not be resolved – treating it as an external dependency
'tty' is imported by commonjs-external-tty, but could not be resolved – treating it as an external dependency
'util' is imported by commonjs-external-util, but could not be resolved – treating it as an external dependency
'fs' is imported by commonjs-external-fs, but could not be resolved – treating it as an external dependency
'net' is imported by commonjs-external-net, but could not be resolved – treating it as an external dependency
Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
No name was provided for external module 'tty' in output.globals – guessing 'tty'
No name was provided for external module 'util' in output.globals – guessing 'util$1'
No name was provided for external module 'fs' in output.globals – guessing 'fs'
No name was provided for external module 'net' in output.globals – guessing 'net'
No name was provided for external module 'buffer' in output.globals – guessing 'buffer'
[...]

PR #651, opened 2020-10-24
Allow external dependecies in custom components

ArcaneDiver:
Hey 👋,
with this PR i update rollup plugins to thier lastest version because plugin like `rollup-plugin-babel` have moved to rollup repo.
Moreover this resolve the problem that a custom react component can't use external dependecies. 

How i resolved the issue: 
- A new version of the [plugin](https://github.com/rollup/plugins/tree/master/packages/commonjs) that resolves commonjs modules allowed us to dont specify `namedExports` in the config
- Add this line to [resolve](https://github.com/rollup/plugins/tree/master/packages/node-resolve/) options mainFields: `['main', 'module', 'jsnext:main']`

Related issues:
#638 #650 #492


wojtek-krysiak:
wow, thanks. I will chceck it out doring the weekend and merge it to beta branch

wojtek-krysiak:
but please make sure that all 4 'yarn bundle' commands from the CI works: they are here https://github.com/SoftwareBrothers/admin-bro/blob/c5cad8d33c365a793547185d97ef8703726a61ac/.github/workflows/push.yml#L58. 

writing this because i see that in the ci there are some messages https://github.com/SoftwareBrothers/admin-bro/runs/1301890895?check_suite_focus=true

wojtek-krysiak:
there is a bundle error in the CI (for some reason github actions passes that - but in the terminal there is some strange output - do you have the same thing on your local dev env?)

image

ArcaneDiver:
I didn't noticed that, probably he escaped me or not. Tomorrow i will investigate.

ArcaneDiver:
After many tests i noticed that the problem don't appear on windows (my os), then i moved to WSL which correctly throws the error. The problem is related to babel because it seem like the bundler can't recognize JSX syntax.

bin/watch-dev.js
```diff
@@ -21,19 +21,20 @@ async function build() {
...
+        'src/backend/utils/filter/filter.ts',
+        'src/backend/decorators/**',
       ],
     },
   })
 }
 
 build().catch((error) => {
   console.log(error)
+  process.exit(1)
```
ArcaneDiver, line 39: With this the process crash with a error code different from 0 and then make the action fail.

bin/watch-dev.js
```diff
@@ -21,19 +21,20 @@ async function build() {
...
-        `${__dirname}/../src/utils/**`,
-        `${__dirname}/../src/backend/utils/view-helpers/view-helpers.ts`,
-        `${__dirname}/../src/backend/utils/filter/filter.ts`,
-        `${__dirname}/../src/backend/decorators/**`,
+        'src/frontend/**',
+        'src/locale/*',
+        'src/utils/**',
+        'src/backend/utils/view-helpers/view-helpers.ts',
+        'src/backend/utils/filter/filter.ts',
+        'src/backend/decorators/**',
```
ArcaneDiver, line 31: Probably new versions of `@rollup/plugin-babel` resolves `include` paths in a different way

ArcaneDiver:
@wojtek-krysiak finally resolved the problem, with b516e9f

wojtek-krysiak:
wow - that is awesome - I will test it in a moment and let's merge this thing and make people happy - with their custom dependencies.

wojtek-krysiak:
I've checked it out today and it looks fine on our test apps. But I would like to spend some time testing it on some real production application. So this is what I will do:

* this/next week we would like to publish v3.3 - we will do this without this PR
* just after that I will merge it and publish it as new beta that we will be able to test it

hope this makes sense. Once again - thanks for your work

wojtek-krysiak (approved):

PR #662, opened 2020-10-28
fix: fix performance issues in large records

dziraf:
This PR fixes the problem where large records (records with a lot of properties/nested properties) would take a very long time to update. At some times it resulted in requests timing out.

The idea was to replace spread assignments with simple assignments (mostly in reduce functions but I also updated it in other places if the change was simple or had a large effect on performance).

With these changes requests which would take 20-30s to complete take up to 1s.


src/backend/adapters/record/base-record.ts
```diff
@@ -112,7 +112,6 @@ class BaseRecord {
    */
   async update(params): Promise<BaseRecord> {
     try {
-      this.storeParams(params)
```
dziraf, line 115: I'm not sure why it was called twice (before and after update). This function takes 100+ms for large records by itself event after refactoring. I decided to remove it because it seemed to work fine + `save` function `storeParams` only after update,

src/backend/decorators/resource/resource-decorator.ts
```diff
@@ -184,11 +184,7 @@ class ResourceDecorator {
       const property = this.properties[propertyName]
 
       const subProperties = flatSubProperties(property)
-      return {
-        ...memo,
-        [propertyName]: property.toJSON(),
-        ...subProperties,
-      }
+      return Object.assign(memo, { [propertyName]: property.toJSON() }, subProperties)
```
dziraf, line 187: I changed it to Object.assign because it's much faster than spread according to this test and it still looks clean: https://www.measurethat.net/Benchmarks/Show/2978/0/lodash-merge-vs-objectassign-vs-spread

`getFlattenProperties` may affect performance because it's used in `toJSON` method which is frequently called everywhere.

wojtek-krysiak (changes requested):
looks good. I added one comment, so please bring that back.

regarding failing tests/linter I originally was thinking about adding this rule to overrides in eslintrc:

```
{
  files: ['src/backend/**/*', 'src/utils/flat/**/*'],
  rules: {
    "no-param-reassign": 'off'
  }
},
```

but let's turn this off globally for all files - it's stupid in case of reduce

The last thing: 100ms is still a lot - please add an issue to GH about this problem - maybe with an example of big record you tested, so we can further iptymise that (for instance flattenProperties could be meoized in the initial phase - but we will have to monitor the memory usage if this big records). I was also thinking about changing reduce to forEach in flat module but not sure if this will help at all.

src/backend/decorators/resource/resource-decorator.ts
wojtek-krysiak: yep - spread creates a new object, object assign doesn't (unless you put {} as the first argument)

src/backend/adapters/record/base-record.ts
```diff
@@ -127,7 +127,6 @@
...
```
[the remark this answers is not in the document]
dziraf: Done

PR #678, opened 2020-11-06
Compose functions for action hooks

Co0sh:
This implements functions for composing `Before` and `After` hooks and `IsFunction` functions. For example part of `ResourceOptions` object, assuming `canManage` is a factory for `Before` hooks and `validatePayload` is a `Before` hook:

```ts
actions: {
  list: {
    before: composeBefores(
      canManage('admin'),
      validatePayload,
    ),
  },
},
```


Co0sh:
Didn't notice it's kind of already implemented.

Issue #672, opened 2020-11-03
overriden SidebarFooter only shows id branding.softwareBrothers is set


TNGPS:
Hello, I spend some time testing the component override, and made my one SidebarFooter , looks like it shows if branding.softwareBrothers is set to true is this a bug or a feature ? 

I would expect to to always show up:

looks like this line is to blame:

https://github.com/SoftwareBrothers/admin-bro/blob/907d03c7ed8f01ca5027655599faabad5c0f6d8c/src/frontend/components/app/sidebar/sidebar.tsx#L61

TNGPS:
instead of removing the conditional, you can consider making is a feature, the only thing that would be needed is to use the same `SideBarFooter` component in the login screen i.s.o the `SoftwareBrothers` component.

E.g. change: 

https://github.com/SoftwareBrothers/admin-bro/blob/907d03c7ed8f01ca5027655599faabad5c0f6d8c/src/frontend/components/login/index.tsx#L126

 into something like this:
```
        {branding.softwareBrothers ? (<SideBarFooter />) : null}
```

Another way to look at this, is to allow the `SoftwareBrothers` component to be overridden.

TNGPS:
I am not sure how to get the two commits related to this issue intro one PR
But  you would need both the changes to resolve this issue,

PR #681, opened 2020-11-10
show branding in sidebar footer only if branding option is set

TNGPS:
relates to issue #672 should be in one PR


TNGPS:
error in the code will create a new one

PR #698, opened 2020-12-02
fix: update appendForceRefresh to append refresh to existing params

dziraf:


src/frontend/components/actions/utils/append-force-refresh.ts
```diff
@@ -9,9 +9,12 @@ export const REFRESH_KEY = 'refresh'
  * @private
  */
 export const appendForceRefresh = (url: string, search?: string): string => {
-  const params = new URLSearchParams(search ?? window.location.search)
-  params.set(REFRESH_KEY, 'true')
-  return `${url}?${params}`
+  const urlObject = new URL(url)
+  const oldParams = search ?? urlObject.search ?? window.location.search
```
dziraf, line 13: I didn't want the new logic to mess up anything in existing apps, so defined `search` argument has precedence over existing url params. In default admin-bro handlers this should only affect the code where response returns `redirectUrl`

SimonB407 (approved):

PR #702, opened 2020-12-08
Allow the ".adminbro" path to be overriden by an env var

MattRoelle:
In order to run adminbro in a serverless environment (aws lambda) this path needs to be overrideable


dziraf:
I'll merge this if you rebase, we've rebranded since you created this PR and both constant name and the default admin dir are different now: https://github.com/SoftwareBrothers/adminjs/blob/master/src/constants.ts#L9

dziraf:
Closing because of lack of response

PR #715, opened 2020-12-15
Fix: Remove error on field when user starts typing or makes a change.

JonaszJestem:
created additional issue to fix those warnings:
https://github.com/SoftwareBrothers/admin-bro/issues/714

### DIFF
--- a/src/frontend/components/property-type/default-type/edit.tsx
+++ b/src/frontend/components/property-type/default-type/edit.tsx
@@ -1,8 +1,9 @@
 /* eslint-disable @typescript-eslint/explicit-function-return-type */
-import React, { FC, useState, memo, useEffect } from 'react'
+/* eslint-disable react-hooks/exhaustive-deps */
+import React, { FC, memo, useEffect, useState } from 'react'
 import Select from 'react-select'
-import { withTheme, DefaultTheme } from 'styled-components'
-import { Input, FormMessage, FormGroup, selectStyles } from '@admin-bro/design-system'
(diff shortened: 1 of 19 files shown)


src/frontend/hooks/use-record/update-record.ts
```diff
@@ -46,6 +46,7 @@ export const updateRecord = (
     populatedModified = true
   }
 
+  delete previousRecord.errors[property]
```
JonaszJestem, line 49: Actual fix

src/frontend/components/property-type/default-type/edit.tsx
```diff
@@ -54,8 +55,18 @@ const TextEdit: FC<CombinedProps> = (props) => {
     if (value !== propValue) {
       setValue(propValue)
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [propValue])
 
+  useEffect(() => {
```
JonaszJestem, line 61: trigger record change when typing in the field with error

SimonB407 (approved):
Two questions. can you describe, why in so many files you disabled `eslint-disable react-hooks/exhaustive-deps`? And also are you sure that `admin-bro/design-system beta 24` is safe? Because I remember we had troubles with some version of beta design system.

Co0sh (approved):

dziraf:
Closing because I'm not sure why `eslint-disable react-hooks/exhaustive-deps` needed to be disabled and this PR is already old

PR #717, opened 2020-12-16
Update edit-action.ts

GuitarWag:
When updating a record the libraries try to run an INSERT SQL command on an existing ID, so, since the record is an instance of himself we don't need to pass the id to the record.update(...newValues) function.


dziraf:
Primary key is not necessarily named `id` plus there might be some cases where a person might actually want to edit the id.

PR #793, opened 2021-03-23
value of availableValues could be string or number

cscovino:
The value property of availabeValues array of PropertyOptions could be either a string or a number, in that way I can use for example typescript's enum to map attributes of a resource.

Now I can use it but without the type declaration of ResourceOptions.


dziraf (approved):

dziraf:
Could you rebase your branch to our `beta` branch?

cscovino:
@dziraf done

Issue #818, opened 2021-04-16
Date fields timezone conversion does not occur on save


andrewg211:
When viewing a datetime field in the Admin UI, the time is converted to the users's local time, however, on save the data is saved exactly as it is entered in the AdminBro UI.

eg. Let's say the following datetime is stored in the DB (UTC):
_2021-04-16 03:00:00_

My timezone is GMT+10, the date renders in the admin UI as expected, +10 hrs:
_2021-04-16 13:00_

Now go ahead edit the time field, let's add just one hour, changing it to _2021-04-16 14:00_

On save, the data is saved in the DB exactly as it is entered, not converted back to UTC - _2021-04-16 14:00:00_

Of course we have a problem now as next time I go to edit the record, the date is converted to my current timezone again adding 10 hours resulting in: 
2021-04-17 00:00

So the conversion is one way, only occurring when loading dates, not saving.

**Installed libraries and their versions**

admin-bro@3.4.0
admin-bro/express@3.1.0
admin-bro/design-system@1.7.0

**To Reproduce**
1. Save a date in the MySQL DB in UTC
2. View the field in the AdminBro UI, you'll see it's converted to you local timezone
3. Change the date and save.
4. Check the DB, the time has not converted back to UTC.

**Expected behavior**
Date should be converted back to UTC on save.

**Desktop (please complete the following information if relevant):**
 - Ubuntu 18.04.1 LTS
 - Browser Chrome

cscovino:
The problem is that the date picker doesn't convert the date to UTC to save it in the DB in the UTC timezone.

cscovino:
I think if it adds a condition in paramsToFormData to check if the value is a date string and then format that date to UTC, it could be the solution, something like:

```
    if(Date.parse(value)){
      const date = new Date(value)
      return formData.set(key, date.toUTCString())
    }
```

https://github.com/SoftwareBrothers/admin-bro/blob/a0d2a17d55af39e03cf3b3f5ab8cd5c09bcaaeb6/src/frontend/hooks/use-record/params-to-form-data.ts#L27-L43

rajeshoodles:
How can I set the default timezone for all dates?

haepeto:
Is there also an option to stick to one (local) timezone only, and save all date and times in that local timezone? That way, when just querying the database, or using another framework to retrieve data, the hours are also correct.

dziraf:
Closing because AdminJS currently saves datetime in ISO format which is standard.

PR #841, opened 2021-05-16
Send date fields to backend in utc timezone

cscovino:
When you create or edit a resource with date properties, the sent payload has the date string in the locale timezone and not in the UTC timezone, so the backend saves the date as if it is in the UTC timezone and does not convert it.

This should fix issue #818


src/frontend/hooks/use-record/params-to-form-data.ts
```diff
@@ -37,6 +37,11 @@ function paramsToFormData(params: Record<string, any>): FormData {
       }
       return formData.set(key, FORM_VALUE_EMPTY_OBJECT)
     }
+    // Convert Date fields to UTC timezone
+    if (Date.parse(value)) {
```
dziraf, line 41: I think we need a different check for Date here because for example `Date.parse(11)` is truthy while it should only convert valid date strings

src/frontend/hooks/use-record/params-to-form-data.ts
```diff
@@ -37,6 +37,11 @@ function paramsToFormData(params: Record<string, any>): FormData {
...
+      const date = new Date(value)
+      return formData.set(key, date.toUTCString())
```
dziraf, line 43: `toISOString` would probably be better since it's a standard format

dziraf (changes requested):

src/frontend/hooks/use-record/params-to-form-data.ts
cscovino: Yes, I agree

src/frontend/hooks/use-record/params-to-form-data.ts
cscovino: Yes, you are right. Let me look for other way.

cscovino: Hi @dziraf, sorry for the late answer. I haven't could find an optimal solution. The solution that I found, without using an external lib, is to use regex to find the date string that returns the date picker. So, as the date picker returns a string with the format `YYYY-MM-DD HH:mm` we could use the conditional:
```if (typeof value === 'string' && value.search(/^([0-9]{4}-[0-9]{2}-[0-9]{2}\s[0-9]{2}:[0-9]{2})$/g) !== -1)```

Other solution is to use moment library with the method `isValid`, for example:
```if (moment(value, 'YYYY-MM-DD HH:mm', true).isValid())```

These solutions, of course, work because I am assuming that the date picker always will return the date in the format `YYYY-MM-DD HH:mm`, so if the format changes, the solutions won't work. That is why I don't like them very much.  

Let me know if you like one of the options or if you found a better solution, so I can update the PR.

Comment (cscovino) [reply src/frontend/hooks/use-record/params-to-form-data.ts:41]: I was viewing the DatePicker component. And actually, the date string always will be in the format `YYYY-MM-DD HH:mm` because admin-bro is formatting the value in the onChange callback:

```
const onDateChange = (newDate: Date): void => {
  if (!disabled) {
    onChange(formatDateProperty(newDate, propertyType))
  }
}
```
https://github.com/SoftwareBrothers/admin-bro-design-system/blob/137b295e0db7ac5bc2827ca897e1e6d546df8e53/src/molecules/date-picker/useDatePicker.ts#L34-38

```
const formatDate = (date: Date): string =&gt; `${date.getFullYear()}-${pad(date.getMonth() + 1)
const formatTime = (date: Date): string => `${pad(date.getHours())}:${pad(date.getMinutes())}`
const formatDateTime = (date: Date): string => `${formatDate(date)} ${formatTime(date)}`
```
https://github.com/SoftwareBrothers/admin-bro-design-system/blob/master/src/utils/date-utils.ts

**Comment by cscovino** [comment]
@dziraf is this issue already fixed?

**Comment by cscovino** [comment]
I don't understand why the tests failed. I run `npm run test` and it passed on my branch.

File: src/frontend/hooks/use-record/params-to-form-data.ts [hunk 3]
```diff
@@ -37,6 +37,16 @@ function paramsToFormData(params: Record<string, any>): FormData {
       }
       return formData.set(key, FORM_VALUE_EMPTY_OBJECT)
     }
+    // Convert Date fields to UTC timezone
+    // Convert Date Picker values (YYYY-MM-DD HH:mm) to UTC timezone
+    // https://github.com/SoftwareBrothers/adminjs-design-system/blob/master/src/utils/date-utils.ts
+    if (
+      value instanceof Date
+      || (typeof value === 'string' && value.search(/^([0-9]{4}-[0-9]{2}-[0-9]{2}\s[0-9]{2}:[0-9]{2})$/g) !== -1)
```
fabiosantoscode, line 45:
```suggestion
      || (typeof value === 'string' && /^[0-9]{4}-[0-9]{2}-[0-9]{2}\s[0-9]{2}:[0-9]{2}$/g.test(value))
```

fabiosantoscode: Maybe a stylistic change, but IIUC using the `test` method is more common.

src/frontend/hooks/use-record/params-to-form-data.ts
fabiosantoscode: I think what @dziraf meant by this, is that a user could feasibly write something date-ish ("YYYY-MM-DD HH:MM") that would be caught by that `if` and turned into an ISO string.

This isn't fixed, right? It seems like if someone types a string that passes this regex, it gets turned into an ISO string.

Instead of checking the string format, perhaps it should only accept `Date` objects here?

src/frontend/hooks/use-record/params-to-form-data.ts
```diff
@@ -37,6 +37,16 @@ function paramsToFormData(params: Record<string, any>): FormData {
...
+    // Convert Date fields to UTC timezone
+    // Convert Date Picker values (YYYY-MM-DD HH:mm) to UTC timezone
+    // https://github.com/SoftwareBrothers/adminjs-design-system/blob/master/src/utils/date-utils.ts
+    if (
+      value instanceof Date
+      || (typeof value === 'string' && /^([0-9]{4}-[0-9]{2}-[0-9]{2}\s[0-9]{2}:[0-9]{2})$/.test(value))
+    ) {
+      const date = new Date(value)
+      return formData.set(key, date.toISOString())
+    }
```
fabiosantoscode, line 49:
```suggestion
    if (value instanceof Date) {
      return formData.set(key, value.toISOString())
    }
```

src/frontend/hooks/use-record/params-to-form-data.ts
fabiosantoscode: Left a suggestion to illustrate

cscovino: Yes, the problem is that the DatePicker returns a string instead of a `Date` object. Maybe the correct way should be to modify the DatePicker component to returns a Date object.

fabiosantoscode: Probably, yes

cscovino: I did the changes, now remains to modify the DatePicker output.

dziraf:
@cscovino I opened a pull request in `adminjs-design-system` repository which updates the Datepicker component to pass a Date object instead of string: https://github.com/SoftwareBrothers/adminjs-design-system/pull/21/files

Could you take a look if using the updated design system would break types in your branch? You can do it by cloning the design system repository, checking out from my branch and running:
```
$ yarn
$ yarn link
$ yarn build && yarn bundle
```
Then on your own branch of core repository:
```
$ yarn link "@adminjs/design-system"
$ yarn dev
```
The `onChange` function had it's arguments types changed from string to Date so it can cause the core repo build to fail

PR #944, opened 2021-11-07
feat: add paramConverter module to convert types before saving a record

dziraf:


.releaserc
```diff
@@ -27,7 +27,7 @@
       "semantic-release-slack-bot",
       {
         "notifyOnSuccess": true,
-        "notifyOnFail": true
+        "notifyOnFail": false
```
dziraf, line 30: no need to notify on slack about failed CI runs

src/backend/actions/new/new-action.ts
```diff
@@ -1,5 +1,6 @@
 import { populator } from '../../utils'
```
plorencrstit, line 1: How about delete action?

plorencrstit (approved):

przemyslaw-szejna-sb (approved):

.releaserc
dziraf: we're using a slack webhook to notify about releases/CI runs (AdminJS Slack), this causes a lot of spam if there is a run which fails

src/backend/actions/new/new-action.ts
dziraf: I don't think we need this for delete action.

The problem which this module is supposed to solve is that AdminJS uses form data to send payload to the backend. Form data's properties are all strings (or booleans, these are actually sent as either true/false which is correct), so even if we use a `number` input in the frontend and type `123` it will be sent as a string: `"123"`.

Usually, ORMs/ODMs do the basic type conversion but it is a problem for AdminJS's `mixed` type because it saves the JSON in the same format it's sent from the frontend.

plorencrstit: Thanks, I got it!

PR #951, opened 2021-11-19
fix: fix bundle function file path resolver

dziraf:


package.json
```diff
@@ -118,7 +118,7 @@
     "@commitlint/cli": "^8.3.5",
     "@commitlint/config-conventional": "^8.3.4",
     "@semantic-release/git": "^9.0.0",
-    "@types/babel-core": "^6.25.6",
+    "@types/babel-core": "^6.25.7",
```
dziraf, line 121: I was getting:
```
node_modules/adminjs/types/src/adminjs-options.interface.d.ts(2,49): error TS2307: Cannot find module 'babel-core' or its corresponding type declarations.
section_end:1637249719:step_script
```
pre-update for some reason

src/adminjs.ts
```diff
@@ -306,7 +306,7 @@ class AdminJS {
     if (src[0] === '/') {
       filePath = src
     } else {
-      filePath = relativeFilePathResolver(src, /Function\.bundle/)
+      filePath = relativeFilePathResolver(src, /.*\.{1}bundle/)
```
dziraf, line 309: This line failed for me because in my case the name wasn't `Function.bundle` but `Proxy.bundle`. The new regex should allow:
- `Function.bundle`
- `Proxy.bundle`
- `Anything.bundle`
- `.bundle`

CC @hank121314

plorencrstit (approved):

src/adminjs.ts
hank121314: Thanks for fixing this issue!
But I am wondering why `Admin.bundle` will convert into `Proxy.bundle` .... 😓

PR #954, opened 2021-11-20
Add branding.companyname to the login title

hexed:


src/frontend/login-template.tsx
```diff
@@ -87,7 +87,7 @@ const html = async (
     <head>
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-      <title>AdminPanel</title>
+      <title>${branding.companyName}</title>
```
dziraf, line 90: can you add a fallback to `AdminPanel` in case `companyName` isn't defined?

dziraf (changes requested):

dziraf (approved):

Issue #937, opened 2021-10-20
Cannot use AdminJS + Typescript + (@types/)React v17


evantahler:
AdminJS relies on react v16.  If the rest of your app uses React v17, Typescipt gets very mad. 

```
../node_modules/@types/react/index.d.ts:3094:14 - error TS2300: Duplicate identifier 'LibraryManagedAttributes'.

3094         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                  ~~~~~~~~~~~~~~~~~~~~~~~~

  ../node_modules/adminjs/node_modules/@types/react/index.d.ts:3091:14
    3091         type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
                      ~~~~~~~~~~~~~~~~~~~~~~~~
    'LibraryManagedAttributes' was also declared here.

../node_modules/@types/react/index.d.ts:3105:13 - error TS2717: Subsequent property declarations must have the same type.  Property 'a' must be of type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>', but here has type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>'.

3105             a: React.DetailedHTMLProps<React.AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>;
                 ~

  ../node_modules/adminjs/node_modules/@types/react/index.d.ts:3104:13
    3104             a: React.DetailedHTMLProps<React.AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>;
                     ~
    'a' was also declared here.
```

This comes from different versions of `@types/react` being installed:

```
$ cat node_modules/adminjs/node_modules/@types/react/package.json | grep version
    "version": "16.14.17",
$ cat node_modules/@types/react/package.json | grep version
    "version": "17.0.30",
```

You can see the miss-matched versions of react with `npm ls react` and that **react-redux** is the culprit:

```
├─┬ @testing-library/react@12.1.1
│ └── react@17.0.2 deduped
├─┬ adminjs@5.2.2
│ ├─┬ @adminjs/design-system@2.0.2
│ │ ├─┬ @carbon/icons-react@10.41.0
│ │ │ └── react@17.0.2 deduped
[...]

PR #955, opened 2021-11-25
Moved @types/react to devDependencies

andiradulescu:
This fixes issue https://github.com/SoftwareBrothers/adminjs/issues/937


dziraf:
Not sure if it's the right way. This will probably cause production bundles to fail since `tsx` files import types

Issue #909, opened 2021-09-06
Add description to fields (fetched via model) on Admin Bro. 


vaibhavmagon:
**Describe the problem feature solves**
At present once you start using the panel on production it gets confusing to remember which fields on the panel do what. As of now we maintain it on confluence (which is not very intuitive).

**Describe the solution you'd like**
Ideally if we can add a (i) button with field names with a small one line description would be really helpful. More like CSS tooltiptext.

**Describe alternatives you've considered**
Having a google sheet/confluence page - non intuitive.

**Acceptance criteria**
Nothing major. This can be fetched as a separate json or can be added as an add on to the field names fetched from Models - mongoose or sequalize.

dziraf:
This is now implemented

PR #969, opened 2021-12-27
feat: add description field

justusburger:
This commit adds the ability to specify a description on the `PropertyOptions` that will be displayed as another `<FormMessage>` below the standard inputs if defined. This makes it easier to provide some more context on fields when the objects become large and nested.

This issue describes more or less the same issue: https://github.com/SoftwareBrothers/adminjs/issues/909

### DIFF
--- a/src/backend/decorators/property/property-options.interface.ts
+++ b/src/backend/decorators/property/property-options.interface.ts
@@ -110,4 +110,6 @@ export default interface PropertyOptions {
    * @new In version 3.3
    */
   reference?: string;
+
+  description?: string;
 }
--- a/src/frontend/components/property-type/boolean/edit.tsx
+++ b/src/frontend/components/property-type/boolean/edit.tsx
@@ -29,6 +29,7 @@ const Edit: React.FC<EditPropertyProps> = (props) => {
         {...property.props}
       />
       <PropertyLabel property={property} props={{ inline: true }} />
+      {property.description && <FormMessage>{property.description}</FormMessage>}
       <FormMessage>{error && error.message}</FormMessage>
     </FormGroup>
   )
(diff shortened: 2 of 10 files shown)


src/backend/decorators/property/property-options.interface.ts
```diff
@@ -110,4 +110,6 @@ export default interface PropertyOptions {
    * @new In version 3.3
    */
   reference?: string;
+
+  description?: string;
```
dziraf, line 114: Can you add a jsdoc description for this field?

src/frontend/interfaces/property-json/property-json.interface.ts
```diff
@@ -47,6 +47,8 @@ export interface PropertyJSON {
    * Property label
    */
   label: string;
+
+  description?: string;
```
dziraf, line 51: Please add a jsdoc comment

src/frontend/components/property-type/boolean/edit.tsx
```diff
@@ -29,6 +29,7 @@ const Edit: React.FC<EditPropertyProps> = (props) => {
         {...property.props}
       />
       <PropertyLabel property={property} props={{ inline: true }} />
+      {property.description && <FormMessage>{property.description}</FormMessage>}
```
dziraf, line 32: I think this should go through i18n to support multi language, might also be a good idea to create a separate component for description.

Another suggestion is to use design system's tooltip component for that: https://storybook.adminjs.co/?path=/story/designsystem-atoms-tooltip--default

and use Carbon Design's `Help` icon as hoverable element: https://www.carbondesignsystem.com/guidelines/icons/library/

dziraf (changes requested):

PR #1015, opened 2022-02-15
fix: fixed array value converting

KS-KIM:
This commit is fixing an issue related to non-string values for Array fields


dziraf (approved):

k-s-k-i-m:
resolve #1016

PR #1021, opened 2022-02-16
WIP - create feature that filter the user logged

daviddossantos:
Hey guys!

I created a feature that filters all data with the base user logged.

Still in WIP, missing to document some things and insert examples.

Feel free to suggestions or improvements.

I felt the need of developing this feature because in some moments was necessary filters all data with the base user logged and does not exist something like that, I believe that this feature will help many peoples


dziraf:
This can actually already be customized via `before` hooks or custom `list` action `handler`. I see no need for this to be included in the core repository, especially that it might not work with all database schemas. An example is that you assume that role would be `admin` for everyone's applications:
https://github.com/SoftwareBrothers/adminjs/pull/1021/files#diff-dcecdf211b25c5195dc11f203b0f5f5e47fb937cc7e46ce031a4937ae8c4c588R50

daviddossantos:
We could change this to some more generical.
> An example is that you assume that role would be admin for everyone's applications.

I did it because when creating a product, for example, was returned categories of other companies that not was of the user logged and not possible customized a `new` action passing a `query`

dziraf:
But you don't need to modify frontend for this.

Example:
```
module.exports = {
  resource: SomeModel,
  options: {
    properties: {
      company: { type: 'reference', reference: 'companies' }
    }
  }
}
```
Company:
```
const filterByUser = async (request, context) => {
  const { currentAdmin } = context;
  request.query = {
    ...request.query,
    'filters.createdByUserId': currentAdmin.id,
  };

  return request;
};

module.exports = {
  resource: SomeResource,
  options: {
    actions: {
      search: {
        before: [filterByUser],
      }
    }
  }
}
```
This will show only Company records where `createdByUserId` is the same as currently logged in user's id.

daviddossantos:
My problem's exactly on the `new` action, because on the field `Categoria` and returned three categories, but the category `frutas` does not belong to the user logged.

image

This `search` action implemented

```
        search: {
          before: async (request, context) => {
            const { currentAdmin } = context
            const { companyId } = currentAdmin
            return {
              ...request,
              query: {
                ...request.params.query,
                'filters.companyId': companyId,
              },
            }
          },
        },
```

daviddossantos:
@dziraf can help me? with this problem

dziraf:
Where did you modify your `search` action? In which resource?

daviddossantos:
on the resource `Product`

> Where did you modify your search action? In which resource?

dziraf:
you want to search categories, so put that in category resource

daviddossantos:
now it's working, I don't believe that it was just that.

oh god, thanks for help @dziraf :)

daviddossantos:
I did this pr without no need, I'm sorry

daviddossantos:
would be interesting to put some reference of as filter by a user logged,  because not it's explicit it

dziraf:
Glad it worked 🎉

PR #1047, opened 2022-03-11
feat: Rebrand cleanup

ariansobczak-rst:
Cleaned up rebranding from SoftwareBrothers

- updated logo 
- changed logo and link in **Need more advanced solutions** section
- replaced SoftwareBrothers "heart" component with MadeWithLove

image


dziraf (approved):
Technically, we don't need to support `softwareBrothers` option anymore if we release this as a breaking change in v6

ariansobczak-rst:
> Technically, we don't need to support `softwareBrothers` option anymore if we release this as a breaking change in v6

Removed all `softwareBrothers` related things. Introduced Migration guide.

PR #1052, opened 2022-03-16
chore: extended dev eslint

ariansobczak-rst:


ariansobczak-rst:
Resigned from global dev linting

PR #1055, opened 2022-03-18
feat: Currency input

ariansobczak-rst:
Requested in https://github.com/SoftwareBrothers/adminjs-design-system/issues/32

Currency input introduced in `design-system` with https://github.com/SoftwareBrothers/adminjs-design-system/pull/35
Demo available on https://github.com/SoftwareBrothers/adminjs-example-app/pull/41
**Example**
```ts
// resource
  options: {
    properties: {
      price: {
        type: "currency",
        props: {
          intlConfig: { locale: "de-DE", currency: "EUR" },
        },
      },
    },
  },
```
**Create**
image
**Filter**
image
**Show**
image
**List**
image

### DIFF
--- a/src/frontend/components/property-type/currency/format-value.ts
+++ b/src/frontend/components/property-type/currency/format-value.ts
@@ -1,0 +1,28 @@
+import { formatCurrencyProperty } from '@adminjs/design-system'
+
+type FormatCurrencyOptions = Parameters<typeof formatCurrencyProperty>[0]
+const optionsKeys: (keyof FormatCurrencyOptions)[] = [
+  'value',
+  'decimalSeparator',
+  'groupSeparator',
+  'disableGroupSeparators',
+  'intlConfig',
+  'decimalScale',
+  'prefix',
+  'suffix',
+]
+
+const pickFormatOptions = (props: Record<string, string>): FormatCurrencyOptions => {
+  const pickedProps = Object.keys(props).reduce((acc, curr) => {
+    if (optionsKeys.includes(curr as any)) acc[curr] = props[curr]
+    return acc
+  }, {} as FormatCurrencyOptions)
+  return pickedProps
+}
+
+const formatValue = (value: string, props: Record<string, string> = {}): string => {
+  const formatOptions = pickFormatOptions({ value: `${value}`, ...props })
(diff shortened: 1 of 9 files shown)


src/frontend/components/property-type/currency/show.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { PureComponent, ReactNode } from 'react'
+import { ValueGroup } from '@adminjs/design-system'
+
+import { EditPropertyProps } from '../base-property-props'
+import formatValue from './format-value'
+
+export default class Show extends PureComponent<EditPropertyProps> {
```
dziraf, line 7: does this have to be a class component for some reason? I also noticed that some other old components are class components but I'm not sure if there's a reason.

src/frontend/components/property-type/currency/format-value.ts
```diff
@@ -0,0 +1,28 @@
...
+const pickFormatOptions = (props: Record<string, string>): FormatCurrencyOptions => {
+  const pickedProps = Object.keys(props).reduce((acc, curr) => {
+    if (optionsKeys.includes(curr as any)) acc[curr] = props[curr]
+    return acc
+  }, {} as FormatCurrencyOptions)
+  return pickedProps
+}
+
+const formatValue = (value: string, props: Record<string, string> = {}): string => {
+  const formatOptions = pickFormatOptions({ value: `${value}`, ...props })
```
dziraf, line 24:
```
value: `${value}`
```
isn't just `value` enough?

dziraf (approved):
can you link the issue which requested this feature into description?

src/frontend/components/property-type/currency/format-value.ts
ariansobczak-rst: Used before to force value as a string, but should value type keep this.

src/frontend/components/property-type/currency/show.tsx
ariansobczak-rst: As you mentioned I created this component based on another one but true should be here consistent and I changed to FC. 
Thanks for pointing that!

dziraf (approved):

PR #1058, opened 2022-03-23
6.0.0

ariansobczak-rst:
# Currency input (https://github.com/SoftwareBrothers/adminjs/pull/1055)
Introduced currency nput as property type
image

# Rebrand cleanup (https://github.com/SoftwareBrothers/adminjs/pull/1047)
Changed logo, removed SoftwareBrothers footer and introduced image

# New rich text editor (https://github.com/SoftwareBrothers/adminjs/pull/1059)
Removed Quill and introduced TipTap rich text editor as the default

image

# Date Picker
Updated DatePicker component to allow to input text via text input. The new Datepicker is included in Design System V3.

# Select
Select component has been moved to Design System (V3), it is now imported from there for reference type property components.

# Phone property (https://github.com/SoftwareBrothers/adminjs/pull/1144)
Added a "phone" property type which is an input with area code and phone text formatting


dziraf:
can we add a list of changes into PR description to more easily keep track of what's going to be added?

src/frontend/components/property-type/currency/edit.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const { onChange, property, record } = props
+  const propValue = record.params?.[property.path] ?? ''
+  const [value, setValue] = useState(propValue)
+  const error = record.errors?.[property.path]
+
+  useEffect(() => {
+    if (value !== propValue) {
+      setValue(propValue)
+    }
+  }, [propValue])
```
RiledUpCrow, line 19: It shouldn't be done like this - it triggers unnecessary rerenders. Instead, split it into two components, keep `propValue` in the outer one, use it as a `key` for the inner one, and move the `value` state to the inner component. This way if the `key` changes, React will create a new instance of that component with fresh state, all within a single rendering pass.

src/frontend/components/property-type/currency/filter.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { PropertyLabel } from '../utils/property-label'
+
+type CurrencyEditPropertyProps = EditPropertyProps & CurrencyInputProps
+
+const Filter: FC<CurrencyEditPropertyProps> = (props) => {
+  const { onChange, property, filter } = props
+
+  const handleChange = useCallback((value) => {
+    onChange(property.path, value)
+  }, [])
```
RiledUpCrow, line 13: This doesn't need memoizing - `CurrencyInput` doesn't use `React.memo` so it will rerender anyway.

src/frontend/components/property-type/richtext/list.tsx
```diff
@@ -1,14 +1,24 @@
-import React from 'react'
+import { Box } from '@adminjs/design-system'
+import truncate from 'lodash/truncate'
+import React, { FC, useEffect, useRef } from 'react'
 import { ShowPropertyProps } from '../base-property-props'
 
-const List: React.FC<ShowPropertyProps> = (props) => {
+const MAX_LENGTH = 15
```
plorencrstit, line 6: can not be changed? should be editable via props IMHO

plorencrstit (approved):

src/frontend/components/property-type/datetime/filter.tsx
```diff
@@ -24,13 +24,13 @@ const Filter: React.FC<FilterPropertyProps> = (props) => {
         <Label>{`- ${translateProperty('from')}: `}</Label>
         <DatePicker
           value={fromValue}
-          onChange={(date: Date): void => onChange(fromKey, date)}
+          onChange={(date): void => onChange(fromKey, date)}
```
RiledUpCrow, line 27: By the way, I think it'd be a great DX improvement if we removed that ESLint rule forcing us to add return types to every function. Typescript has type inference for a reason and typing function return types is only recommended on functions that need to expose a very specific API. An anonymous `onChange` handler doesn't _need_ to return `void`.

src/frontend/components/property-type/richtext/show.tsx
```diff
@@ -1,31 +1,25 @@
...
-    const { property, record } = this.props
-    const value = record.params[property.path]
-    this.contentRef.current.innerHTML = value
-  }
+  useEffect(() => {
+    if (contentRef.current) {
+      const value: string = record.params[property.path] || ''
+      contentRef.current.innerHTML = value
+    }
+  }, [])
```
RiledUpCrow, line 14: Use `dangerouslySetInnerHTML` instead. Would be nice to pass it through some anti-xss tool like [`xss`](https://www.npmjs.com/package/xss).

RiledUpCrow (changes requested):

src/frontend/components/property-type/currency/filter.tsx
ariansobczak-rst: Removed

src/frontend/components/property-type/datetime/filter.tsx
ariansobczak-rst: Good point. Added rule to eslint and removed redundant return types.

dziraf (approved):

PR #1062, opened 2022-03-29
feat: Select

ariansobczak-rst:
Moved select component to design system https://github.com/SoftwareBrothers/adminjs-design-system/pull/39

Replaced in default property with `availableValues` and `reference` select. Changed also in filter components.


src/backend/bundler/config.js
```diff
@@ -44,9 +41,6 @@ const globals = {
   axios: 'axios',
   flat: 'flat',
   recharts: 'Recharts',
-  'react-select': 'ReactSelect',
-  'react-select/async': 'ReactSelectAsync',
-  'react-select/creatable': 'ReactSelectCreatable',
```
dziraf, line 49: Are you sure we don't need these?

ariansobczak-rst: Moved to globals in design system https://github.com/SoftwareBrothers/adminjs-design-system/pull/39/commits/0ee3ceeed3d34ad608526a2f3a924d827fdbd6c2

PR #1064, opened 2022-03-31
feat: allow to override Login component

dziraf:


dziraf:
Closing, this solution wouldn't work. AdminJS v7 will have an easier way of overriding all components.

PR #1078, opened 2022-04-07
fix: fix edit/new submit action recognition

dziraf:


src/frontend/hooks/use-record/use-record.tsx
```diff
@@ -86,7 +86,7 @@ export const useRecord = (
       headers: { 'Content-Type': 'multipart/form-data' },
     }
 
-    const promise = record.id
```
dziraf, line 89: `record` is updated dynamically when you edit the component. If in a specific case, the `id` field is editable, e. g. it's a `passportNumber` where you want to actually type it yourself, this would cause the action to send `edit` request instead of `new`

ariansobczak-rst (approved):

PR #1135, opened 2022-05-12
feat: add default per page setting

RiledUpCrow:


src/adminjs-options.interface.ts
```diff
@@ -242,8 +242,17 @@ export interface AdminJSOptions {
...
    */
   bundler?: BundlerOptions;
+
+  /**
+   * Additional settings.
+   */
+  settings?: Partial<AdminJsSettings>;
 }
 
+export type AdminJsSettings = {
```
dziraf, line 252: I would use `AdminJSSettings` to match vs `AdminJSOptions` (capitalized `JS`)

RiledUpCrow: Done 👍

Issue #535, opened 2020-08-06
Locale should be dynamic


wojtek-krysiak:
**Describe the problem feature solves**

Right now the current locale is set globally, which prevents from, for instance, having multiple languages in one app.

**Describe the solution you'd like**
Locale (only country code) should be changeable when page loads based on the currently logged in user.

dziraf:
Locale is now handled by the frontend and scoped to user

PR #1151, opened 2022-06-10
Feat/language selector continued

jwasiak:


src/adminjs.ts
```diff
@@ -1,27 +1,40 @@
-import * as _ from 'lodash'
 import * as path from 'path'
 import * as fs from 'fs'
 import i18n, { i18n as I18n } from 'i18next'
+/* import isArray from 'lodash/isArray' */
```
dziraf, line 4: please remove this line if `isArray` isn't used

src/frontend/components/app/language-select/language-select.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const LanguageSelect: FC = () => {
+  const handleButton = useCallback((lang) => {
+    window.localStorage.setItem('locale', JSON.stringify(locales[lang]))
+    window.location.reload(true)
+  }, [])
+
+  return (
+    <Box flex>
+      <Button onClick={(): void => handleButton('pl')}>PL</Button>
```
dziraf, line 13: I think this shouldn't be hardcoded, available languages should be chosen from some kind of configuration, preferably `locale.availableLanguages` in `AdminJSOptions`. You can then save them in redux store in `initializeStore` function.  The store setup function would have an additional requirement which checks whether languages are actually existing.
The reason for making this configurable is that if people are creating custom translations and they may not want to translate everything to every possible language.

Explicit return type `: void` should also be unnecessary since React can derive that

src/frontend/login-template.tsx
```diff
@@ -48,7 +48,7 @@ const html = async (
 
   store.dispatch(initializeBranding(branding))
   store.dispatch(initializeAssets(assets))
-  store.dispatch(initializeLocale(admin.locale))
+  store.dispatch(initializeLocale(admin))
```
dziraf, line 51: Why did you chance `admin.locale` to `admin`? 🤔

src/locale/pl.ts
```diff
@@ -0,0 +1,107 @@
...
+    resetFilter: 'Wyczyść',
+    confirmRemovalMany: 'Potwierdź usunięcie {{count}} pozycji',
+    confirmRemovalMany_plural: 'Potwierdź usunięcie {{count}} pozycji',
+    logout: 'Log out',
+    login: 'Log in',
+    seeTheDocumentation: 'See: <1>the documentation</1>',
+    createFirstRecord: 'Utwórz pierwszy wpis',
+  },
+  labels: {
+    navigation: 'Navigation',
```
dziraf, line 27: The translation is only partial, not sure if we want to commit it

src/locale/pt-br.ts
```diff
@@ -0,0 +1,94 @@
+/* eslint-disable @typescript-eslint/camelcase */
+/* spellchecker: disable */
+
+const translations = {
+  actions: {
+    new: 'Criar novo(a)',
```
dziraf, line 6: Brazilian Portuguese locale is already merged, why do you have this in your PR? 🤔

dziraf (changes requested):

dziraf:
Related issue: https://github.com/SoftwareBrothers/adminjs/issues/535

src/adminjs.ts
```diff
@@ -98,26 +115,38 @@ class AdminJS {
...
+      this,
+      global.RegisteredAdapters || [],
+    )
     this.resources = resourcesFactory.buildResources({ databases, resources })
   }
 
   initI18n(): void {
     const language = this.options.locale?.language || locales.en.language
     const defaultTranslations = locales[language]?.translations || locales.en.translations
+    const availableLanguages = locales[language]?.availableLanguages || ['en']
```
dziraf, line 135: I'd change `['en']` to `locales.en.language` so that it isn't hardcoded

src/frontend/components/app/language-select/language-select.tsx
```diff
@@ -0,0 +1,38 @@
...
+const LanguageButtons: FC = () => {
+  const { locale } = useSelector((state: ReduxState) => state)
+  const { availableLanguages } = locale
+  const handleButton = (lng: string): void => {
+    const _locale = { ...locales[lng] }
+    _locale.availableLanguages = availableLanguages
+    window.localStorage.setItem('locale', JSON.stringify(_locale))
+    window.location.reload(true)
+  }
+  if (availableLanguages.length <= 1) {
```
dziraf, line 17: `if (!availableLanguages.length)` should be better

dziraf (changes requested):

dziraf:
Closing for now, will be continued in a separate repository after v7

PR #1154, opened 2022-06-15
Resolve v6 cr issues 

ariansobczak-rst:


dziraf:
@ariansobczak-rst @jwasiak is this ready for review?

.eslintrc.js
```diff
@@ -22,6 +22,9 @@ module.exports = {
       'error',
       2,
     ],
+    '@typescript-eslint/explicit-function-return-type': {
+      allowExpressions: true,
+    },
```
RiledUpCrow, line 27: Personally I always disable this rule completely. A linter can't know when a function needs or doesn't need a stable interface (which is what explicit return type ensures), and type inference is there to make our lives easier.

RiledUpCrow (approved):

PR #1201, opened 2022-08-01
update react dependencies

dziraf:
Updated React dependencies to work properly with React v18:

* Updated `react-redux` to v8 - this shouldn't cause any breaking changes.
* Updated `react-router` and `react-router-dom` to v6. This shouldn't cause any visible changes to most of the applications based on AdminJS. However, if you're importing `buildActionClickHandler`, the `push` param has been replaced with `navigate` due to `useHistory` being removed from `react-router`:

**OLD:**
```
import { useHistory } from 'react-router'
...
const history = useHistory()
const actionClickHandler = buildActionClickHandler({
  ...,
  push: history.push,
})
```
**NEW:**
```
import { useNavigate } from 'react-router'
...
const navigate = useNavigate()
const actionClickHandler = buildActionClickHandler({
  ...,
  navigate,
})
```
You can read more in the migration guide: https://reactrouter.com/docs/en/v6/upgrading/v5
* `react-beautiful-dnd` has been replaced with `react-dnd`. This library has been used for allowing you to change the order of array elements. `react-beautiful-dnd` hasn't been updated to React 18 though and it's development is stale. There shouldn't be any breaking changes caused by this update.


bin/bundle-globals.js
```diff
@@ -53,7 +53,6 @@ const run = async () => {
       react: 'React',
       redux: 'Redux',
       axios: 'axios',
-      recharts: 'Recharts',
```
dziraf, line 56: we've never used it

src/adminjs.ts
```diff
@@ -157,12 +157,15 @@ class AdminJS {
     if (!Database || !Resource) {
       throw new Error('Adapter has to have both Database and Resource')
     }
+
+    // TODO: check if this is actually valid because "isAdapterFor" is always defined.
     // checking if both Database and Resource have at least isAdapterFor method
+    // @ts-ignore
     if (Database.isAdapterFor && Resource.isAdapterFor) {
```
dziraf, line 164: This line has been like this forever but I wonder if it's ever evaluated to false.

`BaseDatabase#isAdapterFor` and `BaseResource#isAdapterFor` are always defined and throw `NotImplementedError` so this check should always return true. We can theoretically make `BaseResource` and `BaseDatabase` abstract classes or we can simply remove this "if" condition but I decided to do it another time. I added `@ts-ignore` because Typescript v3 ignored this line while Typescript v4 says it always returns true.

src/frontend/components/app/sort-link.tsx
```diff
@@ -13,33 +13,32 @@ export type SortLinkProps = {
...
-    )
-  }
+const SortLink: React.FC<SortLinkProps> = (props) => {
+  const { sortBy, property, direction } = props
+  const location = useLocation()
+
+  const isActive = useMemo(() => sortBy === property.propertyPath, [sortBy, property])
+
+  const query = new URLSearchParams(location.search)
+  const oppositeDirection = (isActive && direction === 'asc') ? 'desc' : 'asc'
```
plorencrstit, line 23: Enum for `desc` and `asc` would be great

plorencrstit (commented):
Looks fine but I don't have any experience in React 18

src/frontend/components/app/sort-link.tsx
dziraf: This is currently used in too many places, I'll focus on rewriting this when working on AdminJS v7, otherwise Jarek and others will have more work to do when synchronizing changes while they split frontend to another repo

PR #1203, opened 2022-08-01
Fix bug:  value is not iterable

alipirpiran:
continue the loop if the value is undefined.

### DIFF
--- a/src/utils/param-converter/convert-nested-param.ts
+++ b/src/utils/param-converter/convert-nested-param.ts
@@ -12,12 +12,10 @@ const convertNestedParam = (
 
   let value = parentValue[path]
 
-  if (type === 'mixed') {
+  if (type === 'mixed' && value) {
     const nestedSubProperties = subProperty.subProperties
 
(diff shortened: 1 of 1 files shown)


src/utils/param-converter/convert-nested-param.ts
```diff
@@ -16,6 +16,8 @@ const convertNestedParam = (
     const nestedSubProperties = subProperty.subProperties
 
     for (const nestedSubProperty of nestedSubProperties) {
+      if (!value) continue;
```
dziraf, line 19: This would simply make it skip all loops, changing the `if` condition to:

```javascript
 if (type === 'mixed' && value) {
```

should make more sense.

dziraf (changes requested):
This would simply make it skip all loops, changing the `if` condition to:

```javascript
 if (type === 'mixed' && value) {
```

should make more sense.

alipirpiran:
yes. your right, value is not changing in the loop.
I fixed it. thanks.

dziraf (approved):

PR #1206, opened 2022-08-02
feat(small): make topbar overidable

BarejaPawel:


.eslintrc.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
     'import/extensions': 'off',
     'mocha/no-hooks-for-single-case': 'off',
     'no-param-reassign': 'off',
+    'max-len': 'off',
```
dziraf, line 33: We shouldn't turn it off globally, most of the time code can be formatted and in other situations you can disable the rule just for one line or block of code.

package.json
```diff
@@ -8,7 +8,7 @@
     "test": "mocha ./spec/index.js",
     "types": "tsc",
     "clean": "rm -rf lib && mkdir lib && rm -fr types && mkdir types",
-    "build": "babel src --out-dir lib --copy-files --extensions '.ts,.js,.jsx,.tsx'",
+    "build": "babel src --out-dir lib --copy-files --extensions .ts,.js,.jsx,.tsx",
```
dziraf, line 11: removing `'` will break the command for unix machines, you can try to change it to:
```
    "build": "babel src --out-dir lib --copy-files --extensions \".ts,.js,.jsx,.tsx\"",
```
otherwise I would just keep the old version

dziraf (changes requested):

dziraf (approved):

PR #1220, opened 2022-08-22
feat: add support for additional text below the sidebar logo

Half-Shot:
This allows us to specify which environment our application is on. It felt like a good idea to make this generic and upstream the work rather than making a custom component.

### DIFF
--- a/src/adminjs-options.interface.ts
+++ b/src/adminjs-options.interface.ts
@@ -349,7 +349,10 @@ export type BrandingOptions = {
    * @new since 6.0.0
    */
   withMadeWithLove?: boolean;
-
+  /**
+   * Optional additional text below the logo/companyName.
+   */
+  additionalText?: string;
   /**
    * URL to a favicon
    */
--- a/src/frontend/components/app/sidebar/sidebar-branding.tsx
+++ b/src/frontend/components/app/sidebar/sidebar-branding.tsx
@@ -43,7 +43,7 @@ const h = new ViewHelpers()
 const SidebarBranding: React.FC<Props> = (props) => {
   const { branding } = props
   const { logo, companyName } = branding
-  return (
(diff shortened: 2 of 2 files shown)


src/frontend/components/app/sidebar/sidebar-branding.tsx
```diff
@@ -55,7 +55,10 @@ const SidebarBranding: React.FC<Props> = (props) => {
         />
       ) : <h1>{companyName}</h1>}
     </StyledLogo>
-  )
+    { branding.additionalText &&
+      <span>{branding.additionalText}</span>
```
dziraf, line 59: I'd use `Text` (`import { Text } from '@adminjs/design-system`). This uses AdminJS global styles.

src/adminjs-options.interface.ts
```diff
@@ -349,7 +349,10 @@ export type BrandingOptions = {
    * @new since 6.0.0
    */
   withMadeWithLove?: boolean;
-
+  /**
+   * Optional additional text below the logo/companyName.
+   */
+  additionalText?: string;
```
dziraf, line 355: How about just adding `NODE_ENV` here:
https://github.com/SoftwareBrothers/adminjs/blob/master/src/frontend/components/app/version.tsx
if it's just to display the environment information?

We can also add additional option to `branding` but I'd use a different name like `logoTitle`, `logoSubtitle`, `headerText`, `headerMessage`, `menuHeaderMessage`, `menuHeaderText` which would indicate where it'll appear just by looking at it's name.

Half-Shot:
Sadly, I've stopped working on the project that used AdminJS and have just not found enough time to come back to this PR. I'm going to close it so as not to clog up the queue, but anyone is free to take it on.

PR #1222, opened 2022-08-23
fix: use @react-forked/dnd for drag and drop

dziraf:
`@react-forked/dnd` is a continuation of `react-beautiful-dnd` which Atlassian stopped developing.

`react-dnd` had various issues.


src/frontend/components/property-type/array/edit.tsx
```diff
@@ -1,155 +1,70 @@
-import React, { MouseEvent, useCallback, useMemo, memo, useRef } from 'react'
+import React, { MouseEvent, useCallback } from 'react'
```
dziraf, line 1: This file is exactly the same as it was pre-6.1.0

plorencrstit (approved):

ariansobczak-rst (approved):

PR #1258, opened 2022-09-20
fixed no display in filter drawer previous boolean and reference values

jwasiak:


src/frontend/components/property-type/reference/filter.tsx
```diff
@@ -5,39 +5,46 @@ import ApiClient from '../../../utils/api-client'
...
+    })
+    return options
   }
 
   render(): ReactNode {
     const { property, filter } = this.props
+    const { options } = this.state
     const value = typeof filter[property.path] === 'undefined' ? '' : filter[property.path]
-    const selected = (this.options || []).find((o) => o.value === value)
+    const selected = (options || []).find((o) => o.value === Number(value))
```
dziraf, line 47: There shouldn't be `Number(value)` here. ID columns can be any type:
- uuid
- ObjectId (in mongo)
- normal string (email, etc)

dziraf (changes requested):

dziraf (approved):

PR #1262, opened 2022-09-23
added isShow to page props

acamara2016:


src/adminjs-options.interface.ts
```diff
@@ -55,19 +55,19 @@ export interface AdminJSOptions {
    * path, under which, AdminJS will be available. Default to `/admin`
    *
    */
-  rootPath?: string;
+  rootPath?: string
```
dziraf, line 58: Why irrelevant changes?

dziraf (changes requested):

src/adminjs-options.interface.ts
```diff
@@ -379,16 +377,21 @@ export type AdminPage = {
...
   /**
    * Page icon
    */
-  icon?: string;
+  icon?: string
+
+  /**
+   * Page visibility
+   */
+  isShown?: boolean
```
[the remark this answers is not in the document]
acamara2016: I changed `isShown` to `isVisible` in the latest commit with type either `boolean | IsFunction`

src/frontend/interfaces/page-json.interface.ts
```diff
@@ -16,4 +19,9 @@ export interface PageJSON {
    * Page icon
    */
   icon?: string;
+
+  /**
+   * Page visibility
+   */
+  isVisible: boolean | IsFunction;
```
dziraf, line 26: `PageJSON` is something returned to the frontend from the backend so it cannot be a function (`IsFunction`) and has to be pre-evaluated.

`pagesToStore` would have to do something like:
```
  Object.entries(pages).map(([key, adminPage]) => {
      let isVisible = true;
      if (adminPage.isVisible) {
         if (typeof adminPage.isVisible === 'function') {
           isVisible = adminPage.isVisible(currentAdmin)
         } else {
           isVisible = adminPage.isVisible
         }
      }
      return {
        name: key,
        component: adminPage.component,
        icon: adminPage.icon,
        isVisible,
      }
  })
```

dziraf (changes requested):

src/frontend/interfaces/page-json.interface.ts
```diff
@@ -1,3 +1,6 @@
+import { IsFunction } from 'src/backend'
```
dziraf, line 1: Sorry I haven't noticed this earlier. Can you change the import path to be relative (not start with `src/`)?

dziraf (changes requested):
One last thing, have you tested if this actually hides the page from the sidebar?

I'm not sure if any changes here are required: https://github.com/SoftwareBrothers/adminjs/blob/5145c48b7eb0758a603a4ee5435aa3883193ddd5/src/frontend/components/app/sidebar/sidebar-pages.tsx

PR #1268, opened 2022-10-05
Feat/add styles ids

jwasiak:


src/frontend/components/actions/bulk-delete.tsx
```diff
@@ -66,16 +66,20 @@ const BulkDelete: React.FC<ActionProps & AddNoticeProps> = (props) => {
...
   }
 
+  const cssTag = `${resource.id}-bulk-delete-content`
+  const tableTag = `${resource.id}-bulk-delete-table`
+  const footerTag = `${resource.id}-bulk-delete-footer`
+
   return (
     <>
-      <DrawerContent>
+      <DrawerContent data-css={cssTag}>
```
dziraf, line 75: Why `cssTag` and not `contentTag`? It's the same in other files, one of the tags is always `cssTag` 🤔

dziraf: + why do you hardcode `bulk-delete` instead of doing `${action.name}`?

src/frontend/components/actions/edit.tsx
```diff
@@ -41,15 +41,20 @@ const Edit: FC<ActionProps> = (props) => {
     return false
   }
 
+  const cssTag = `${resource.id}-${action.name}-drawer-content`
```
dziraf, line 44: I would unify naming a bit.

`DrawerContent`'s `data-css` is named:
`User-edit-drawer-content` in this file but for bulk delete it's just
`User-bulk-delete-content`.

I think the best way would be to create:
`src/frontend/utils/css-tags.ts`

with a set of functions, for example:

```
export const getActionElementCss = (resource: ResourceJSON, action: ActionJSON, suffix: string) => `${resource.id}-${action.name}-${suffix}`
```

Usage:
```
const drawerContentTag = getActionElementCss(resource, action, 'drawer-content')
```

src/frontend/components/actions/list.tsx
```diff
@@ -54,8 +54,10 @@ const List: React.FC<ActionProps> = ({ resource, setTag }) => {
     navigate({ search: search.toString() })
   }
 
+  const cssTag = `${resource.id}-list-action-content`
```
dziraf, line 57: Same here, you should also be able to use `action.name` to get the name of the action

dziraf (changes requested):

src/frontend/components/actions/bulk-delete.tsx
jwasiak: done

src/frontend/components/actions/edit.tsx
jwasiak: done

src/frontend/components/actions/list.tsx
jwasiak: done

src/frontend/components/actions/edit.tsx
```diff
@@ -41,15 +42,20 @@ const Edit: FC<ActionProps> = (props) => {
     return false
   }
 
+  const contentTag = getActionElementCss(resource, action, 'drawer-content')
+  const formTag = getActionElementCss(resource, action, 'form')
+  const footerTag = getActionElementCss(resource, action, 'drawer-footer')
```
dziraf, line 47: As I said, unify naming. You use `content` here:
https://github.com/SoftwareBrothers/adminjs/pull/1268/files#diff-33aa50ef73bc5baf6e4a148d08aa9adf1d949ab9f82674ca4e72e4da7264c623R69
and in this component it's `drawer-content`. I think `drawer-content` is a better name since it matches the component

dziraf (changes requested):

PR #1269, opened 2022-10-06
Fix/drawer boolean values

jwasiak:


src/frontend/components/property-type/reference/filter.tsx
```diff
@@ -5,39 +5,46 @@ import ApiClient from '../../../utils/api-client'
...
+    }
   }
 
   handleChange(selected: SelectRecord): void {
     const { onChange, property } = this.props
     onChange(property.path, selected ? selected.value : '')
   }
 
-  async loadOptions(inputValue: string): Promise<Array<{value: string; label: string }>> {
+  async loadOptions(inputValue: string): Promise<Array<{value: string | number; label: string }>> {
```
dziraf, line 30: `Array<SelectRecord>`>

dziraf (changes requested):

src/frontend/components/property-type/reference/filter.tsx
jwasiak: fixed

dziraf:
@jwasiak there are conflicts with `master` branch

PR #1271, opened 2022-10-07
feat: add key-value property type

dziraf:
https://user-images.githubusercontent.com/16668924/194540794-0054bdd2-e667-482f-a008-1e746fed3132.mov

### DIFF
--- a/src/frontend/components/property-type/base-property-component.tsx
+++ b/src/frontend/components/property-type/base-property-component.tsx
@@ -6,6 +6,7 @@ import ErrorBoundary from '../app/error-boundary'
 
 import * as ArrayType from './array'
 import * as MixedType from './mixed'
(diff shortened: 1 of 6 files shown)


src/frontend/components/property-type/base-property-component.tsx
```diff
@@ -108,6 +111,17 @@ const BasePropertyComponent: React.FC<BasePropertyComponentProps> = (props) => {
     )
   }
 
+  if (baseProperty.type === 'key-value') {
```
plorencrstit, line 114: Would not be safer to use `enum` here? Generally to refactor `PropertyType` from `type` to `enum`?

plorencrstit (approved):

ariansobczak-rst (approved):

PR #1273, opened 2022-10-10
feat: replace AdminJS.bundle with ComponentLoader

RiledUpCrow:


packages/adminjs-e2e/src/admin.options.js
```diff
@@ -1,4 +1,5 @@
 const { default: AdminJS } = require('adminjs')
+const ComponentLoader = require('adminjs/utils/component-loader')
```
dziraf, line 2: Let's just remove the entire `adminjs-e2e` folder since it's never been officially released AFAIK and it's not updated after v4.

src/cli.ts
```diff
@@ -15,7 +15,7 @@ program.version(pkg.version)
 program
   .command('bundle <configFile>')
   .description([
-    'Bundle all custom components addde by using AdminJS.bundle(filePath).',
+    'Bundle all custom components addded by using ComponentsLoader',
```
dziraf, line 18: `addded` -> `added`

src/utils/component-loader.ts
```diff
@@ -0,0 +1,165 @@
...
+export interface ComponentDetails {
+  overrides: boolean
+  filePath: string
+}
+
+export default class ComponentLoader {
+  protected components: Record<string, ComponentDetails>
+
+  public constructor(protected readonly themeId?: string) {
+    this.components = {}
```
dziraf, line 16: Let's also add `this.themeId = themeId ?? null`

src/utils/component-loader.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
```
dziraf, line 18: I think we will need a way to merge multiple loaders for example when using features like `@adminjs/passwords`.

How about we add something like:

```
public mergeLoader(loader: ComponentLoader) {
   if (this.getThemeId() || loader.getThemeId()) {
     throw new Error('Cannot merge Theme loaders');
   }

   this.components = {
       ...loader.getComponents(),
       ...this.getComponents(),
   }
}
```

dziraf (changes requested):

src/utils/component-loader.ts
dziraf: Feature packages would also export their loaders which you'd have to merge with you own loader.

RiledUpCrow: I don't like that approach, I'd rather share a single loader if it's possible with the current API. I'll check the code and see if I can come up with something.

dziraf (approved):

PR #1288, opened 2022-10-27
feat: allow searching by non-title properties

RiledUpCrow:


jwasiak (approved):

src/frontend/utils/api-client.ts
```diff
@@ -137,16 +137,23 @@ class ApiClient {
    * @param   {Object}  options
    * @param   {String}  options.resourceId  id of a {@link ResourceJSON}
    * @param   {String}  options.query       query string
+   * @param   {String}  options.property    optional property name
    *
    * @return  {Promise<SearchResponse>}
    */
-  async searchRecords({ resourceId, query }: {
+  async searchRecords({ resourceId, query, property }: {
```
dziraf, line 144: This is ok but can we name it `searchProperty` or `searchBy`?

RiledUpCrow: Sure, good catch 😄

dziraf (approved):

PR #1356, opened 2022-12-24
refactor: fix <TextEdit /> bugs

AshotN:
`<TextEdit />` currently has a few issue, and it's something that causes a lot of issues for me

1. If the user has a field which the value of is currently `null`, there is no way for the user to know null vs blank.
2. If the current field is `null` and during an edit the user focuses the field but makes no changes, during a save the client will tell the server that the value is `""` for strings or `0` for numbers. This is an issue where an edit is made without the user even knowing it happened, this can cause validation errors and/or other issues.
3. The component is overly complex and uses odd event listeners like `onBlur` and `onKeyDown` which is unnecessary.
4. The `useEffect` seemingly does nothing 

This PR contains a fix to `<TextEdit />`

But in addition has a feature proposal that I haven't ironed out yet. I propose to add some form of a reset for the field, it would make it clear to the user that the field has been edited. I have the functionality working, but the styling is pretty bad.

This video showcases the fix for `<TextEdit />` as you are able to focus and unfocus without value change.
The "Reset" text will only show for a field that has been modified. 

The "Price per" field is actually not fixed since currency type doesn't use the same `<TextEdit />` component, so the old behavior is evident there.

recording

### DIFF
--- a/src/frontend/components/property-type/base-property-component.tsx
+++ b/src/frontend/components/property-type/base-property-component.tsx
@@ -1,6 +1,6 @@
-import { Box } from '@adminjs/design-system'
+import { Box, Label } from '@adminjs/design-system'
 import { ReactComponentLike } from 'prop-types'
-import React, { useMemo } from 'react'
+import React, { useMemo, useState } from 'react'
 
 import ErrorBoundary from '../app/error-boundary'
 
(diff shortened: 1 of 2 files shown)


AshotN:
image

Perhaps buttons instead would look better?

PR #1357, opened 2022-12-27
feat: include context when calling adapter functions

### LINKED ISSUE #1353: currentAdmin in BaseResource
**Describe the problem feature solves**
I am making a custom adapter that connects to another server, and I want to be able to pass the currently authenticated user so the external server can do it's auth checks.

**Describe the solution you'd like**
Allow currentAdmin to be accessible from BaseResource

**Acceptance criteria**
User information being accessible from adapters

AshotN:
Closes #1353

Is backwards compatible, adapters do not need to use `currentAdmin` if they don't need it

### DIFF
--- a/src/backend/adapters/record/base-record.ts
+++ b/src/backend/adapters/record/base-record.ts
@@ -133,10 +133,10 @@ class BaseRecord {
    * @param  {object} params all field with values which has to be updated
    * @return {Promise<BaseRecord>}        given record (this)
    */
-  async update(params): Promise<BaseRecord> {
+  async update(params, currentAdmin?: CurrentAdmin): Promise<BaseRecord> {
     try {
       this.storeParams(params)
-      const returnedParams = await this.resource.update(this.id(), params)
+      const returnedParams = await this.resource.update(this.id(), params, currentAdmin)
       this.storeParams(returnedParams)
     } catch (e) {
       if (e instanceof ValidationError) {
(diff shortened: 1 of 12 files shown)


dziraf:
@AshotN How about we just pass `context` as the last argument? This way you'd be able to access even more action/request scoped details.

AshotN:
That sounds good, I'll edit the PR

AshotN:
I had not originally noticed that AdminJS itself relies on a lot of the adapter functions, so those need to pass the context as well. I made the changes, but the tests are failing and I'm not sure how to fix them. Would like some help on that front

dziraf:
The tests aren't failing but there are some type errors.

We could try to type it as `context?: Partial<ActionContext> = {}`. It should make the third argument optional but it will be up to adapter's authors to make sure that `context.currentAdmin` or other attributes are available. This shouldn't break anything since none of current adapters use `context`.

AshotN:
Ya I just realized that without having context as optional, current adapters will type-error. I will make it optional again

AshotN:
@dziraf Changed back to optional, please review and let me know if it seems ok to you.

src/backend/adapters/record/base-record.ts
```diff
@@ -135,7 +135,7 @@ class BaseRecord {
    * @param  {ActionContext}           context
```
dziraf, line 135: `context` -> `[context]`

This should make jsdoc treat context as optional.

dziraf (changes requested):
@AshotN Do you think this is ready to be merged otherwise?

AshotN:
I believe so. I don't see any glaring issues but I will test it with my adapter again.

AshotN:
@dziraf I think this is good to go

dziraf (approved):
This looks good to me, worst case scenario we will revert it.

PR #1380, opened 2023-01-14
fix <TextEdit /> bugs

AshotN:
This is a re-attempt at https://github.com/SoftwareBrothers/adminjs/pull/1356


src/frontend/components/property-type/array/edit.tsx
```diff
@@ -20,14 +20,12 @@ type ItemRendererProps = {
...
 
 const ItemRenderer: React.FC<EditProps & ItemRendererProps> = (props) => {
-  const { ItemComponent, property, onDelete, index, record, isDraggable } = props
-  const uniqueDraggableId = window.btoa(unescape(encodeURIComponent(`${JSON.stringify(flat.get(record.params, property.path))}-${property.path}`)))
+  const { ItemComponent, property, onDelete, index, isDraggable } = props
 
   return (
     <Draggable
-      draggableId={uniqueDraggableId}
+      draggableId={property.path}
```
dziraf, line 27: Does this fix the issue? `property.path` is still not unique as it changes when you drag an element between indices.

dziraf (commented):
@AshotN I think this PR is missing the `TextEdit` changes as we reverted them? Refering to PR title.

src/frontend/components/property-type/array/edit.tsx
AshotN: That's a fair point will have to dive into that.

AshotN:
You're right will add that back into the PR

src/frontend/components/property-type/array/edit.tsx
AshotN: There doesn't seem to be a convenient value to use as a unique key.

dziraf: One solution I thought of would be to create a "virtual" property `draggableId` for all existing array elements when the HOC component first loads, then assign it also when you add new elements. We could use some kind of library like `nanoid` to create a unique identifier. There could be an issue where AdminJS also sends this property in request payload and an adapter tries to save it. AFAIK some ORMs skip attributes that are not defined in models but other ones could throw errors. A solution to this could be that we treat all properties prefixed with double underscore (`__typename`, `__draggableId`, etc) as virtual/readonly and remove them when form is submitted.

AshotN: Would you want that solution as part of this PR?

dziraf: Ideally, yes. If you have doubts though and `property.path` change works exactly the same as `uniqueDraggableId` (as it was before) then it's fine to do it another time.

AshotN: I would prefer handling it in a separate PR, the current implementation has some quirks like focus going to the wrong field when dragging elements. But that is an existing issue.

AshotN:
Currently AdminJS has a lot of different components for different types of data, they have repeating functionality. This PR doesn't address all of them. If the changes so far seem sound, I can continue to try to implement the same for the rest of the property-type components.

PR #1383, opened 2023-01-17
feat: extended return type of Before action by ActionRequest

VeryVicious:
Motivation behind this change is fact, that `Before` action return type strictly enforces  to use async. 
It would be great to  also enable returning `ActionRequest` synchronously.


dziraf:
Hm, let's make the same change to `After`?

dziraf (approved):

PR #1403, opened 2023-02-04
feat: importing themes

RiledUpCrow:
# Importing Themes

Themes let you change the style and default components of your admin panel. There are three elements added to the app by each theme:

- component overrides - React components that replace a subset of [overridable components](https://github.com/SoftwareBrothers/adminjs/blob/master/src/frontend/utils/overridable-component.ts)
- configuration for [`@adminjs/design-system`](https://github.com/SoftwareBrothers/adminjs-design-system) - this is merged into the default config, and then user's branding is applied on top
- custom styles - a single global CSS file with additional theme styles

## Setup

You can add `availableThemes` configuration option with an array of theme objects (for example imported from [`@adminjs/themes`](https://www.npmjs.com/package/@adminjs/themes) package). The first theme on the list will be active by default. This can be changed with `defaultTheme` option - specify the ID of a theme - these are usually same as the named exports.

```ts
import AdminJS from 'adminjs'
import { light, dark } from '@adminjs/themes'

new AdminJS({
  availableThemes: [light, dark],
  defaultTheme: 'dark', // same as `dark` export above
})
```
## Per-user themes

Additionally theme can be configured per-user by returning theme ID in the `theme` property of the `CurrentAdmin` object - this should be implemented via the framework plugin, i.e. `authenticate` function in [`@adminjs/express`](https://www.npmjs.com/package/@adminjs/express). Since you have full control of that object, returned theme can be selected based on things like user's role or a property in the database.

There's no UI control for changing themes, but you can implement this yourself. The general idea is to keep theme ID in the user resource, and add a button that will update the current user record with the selected theme. Alternatively you can let the user edit their own record and offer a custom select property. Note that for the `currentAdmin` object to update the user most likely needs to sign in again.

### DIFF
--- a/src/frontend/hoc/allow-override.tsx
+++ b/src/frontend/hoc/allow-override.tsx
@@ -29,7 +29,7 @@ function allowOverride<P extends Record<string, unknown>>(
      */
     if (typeof window !== 'undefined') {
       Component = window.AdminJS?.UserComponents?.[name]
-        ?? (window as any).THEME?.Components?.[name]
+        ?? (window as any).THEME_COMPONENTS?.[name]
         ?? OriginalComponent
     }
 
--- a/src/adminjs-options.interface.ts
+++ b/src/adminjs-options.interface.ts
@@ -251,6 +251,25 @@ export interface AdminJSOptions {
    * Additional settings.
    */
   settings?: Partial<AdminJSSettings>;
+
+  /**
+   * List of available themes, for example exports of the `@adminjs/themes` npm package.
+   */
+  availableThemes?: ThemeConfig[];
+
+  /**
+   * ID of the default theme. If not provided, the first theme from the `availableThemes`
+   * list will be used.
+   */
+  defaultTheme?: string;
+}
+
+export type ThemeConfig = {
+  id: string,
+  name: string,
+  data: Partial<ThemeOverride>;
+  bundlePath: string;
+  stylePath: string;
 }
 
 export type AdminJSSettings = {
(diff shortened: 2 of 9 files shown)


ariansobczak-rst (approved):

src/frontend/hoc/allow-override.tsx
```diff
@@ -29,8 +29,8 @@ function allowOverride<P extends Record<string, unknown>>(
      *
      */
     if (typeof window !== 'undefined') {
-      Component = (window as any).AdminJS?.UserComponents?.[name]
-        ?? (window as any).THEME?.Components?.[name]
+      Component = window.AdminJS?.UserComponents?.[name]
+        ?? window.THEME_COMPONENTS?.[name]
```
dziraf, line 33:
```
Property 'THEME_COMPONENTS' does not exist on type 'Window & typeof globalThis'.
```
there seem to be some issues with global namespace

src/frontend/bundle-entry.jsx
```diff
@@ -26,7 +26,7 @@ const { locale } = store.getState()
 const { i18n } = initTranslations(locale)
 
 const Application = (
-  <Suspense fallback="...is loading">
+  <Suspense fallback="AdminJS is loading">
```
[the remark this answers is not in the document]
ariansobczak-rst: Added <AppLoader /> component

[Screencast from 2023-04-04 10-07-33.webm](https://user-images.githubusercontent.com/85543813/229731253-a513343f-abe7-46f8-8e12-8042493a0efa.webm)

dziraf (approved):

src/frontend/hoc/allow-override.tsx
ariansobczak-rst: Restored window as any to pass tests

PR #1414, opened 2023-02-14
fix: currency field error in list

jwasiak:


src/frontend/components/property-type/currency/format-value.ts
```diff
@@ -14,7 +14,11 @@ const optionsKeys: (keyof FormatCurrencyOptions)[] = [
 
 const pickFormatOptions = (props: Record<string, string>): FormatCurrencyOptions => {
   const pickedProps = Object.keys(props).reduce((acc, curr) => {
-    if (optionsKeys.includes(curr as any)) acc[curr] = props[curr]
+    if (optionsKeys.includes(curr as any)) {
+      if (props[curr] !== null) {
+        acc[curr] = props[curr].toString()
```
dziraf, line 19: This will throw an error if `props[curr]` is `undefined`. If you do `String(props[curr])` it will also not work because it will convert to `"undefined"` string. I think we need an additional check for `undefined` for safety

dziraf (changes requested):

src/frontend/components/property-type/currency/format-value.ts
jwasiak: added

dziraf (approved):

PR #1421, opened 2023-02-17
fix: translate boolean values on filter

DiogoAbu:
The filter component of boolean property was missing translation.


dziraf (approved):

dziraf:
The target branch should be `master` since this is not a breaking change

PR #1424, opened 2023-02-20
fix: translate phone input placeholder and message

DiogoAbu:
The edit component of phone property was missing translations.


src/frontend/components/property-type/phone/edit.tsx
```diff
@@ -27,6 +31,8 @@ const Edit: FC<EditPropertyProps> = (props) => {
           name: property.path,
           required: property.isRequired,
         }}
+        searchPlaceholder={translateLabel('search', resource.id)}
```
dziraf, line 34: Let's use `translateMessage` for placeholder too, `translateLabel` is mostly used for translating some headers.

dziraf (changes requested):

src/frontend/components/property-type/phone/edit.tsx
dziraf: Could you also add translations to the core translations for all languages? We usually just use https://www.deepl.com/en/translator for this

PR #1426, opened 2023-02-21
feat: Record relations

ariansobczak-rst:


ariansobczak-rst:
Implemented in https://github.com/SoftwareBrothers/adminjs-relations

PR #1428, opened 2023-02-21
feat: useModal hook can handle all Modal props

jwasiak:


src/frontend/hooks/use-modal.ts
```diff
@@ -1,54 +1,65 @@
...
+    const { modalProps, type, confirmAction, resourceId } = props
+    let buttons: ModalProps['buttons'] = modalProps.buttons || []
 
     const handleConfirm = () => {
       closeModal()
       if (confirmAction) confirmAction()
     }
 
+    if (buttons.length) {
+      const btns = buttons.map((button) => {
```
ariansobczak-rst, line 26: **suggestion:** Reasign to temp is redundant here 

```suggestion
      buttons = buttons.map((button) => {
```

ariansobczak-rst (approved):

PR #1431, opened 2023-02-23
fix: initial locale problem after changing languages

jwasiak:


src/frontend/bundle-entry.jsx
```diff
@@ -21,8 +21,14 @@ const env = {
 
 const store = createStore(window.REDUX_STATE)
 const theme = window.THEME
+const adminLocale = window.REDUX_STATE.locale
```
dziraf, line 24: Is it safe to access `window.REDUX_STATE` instead of `useSelector`?

CC @ariansobczak-rst

src/frontend/bundle-entry.jsx
```diff
@@ -21,8 +21,14 @@ const env = {
...
 const currentLocale = JSON.parse(window.localStorage.getItem('locale'))
-const locale = currentLocale || window.REDUX_STATE.locale
+let locale = null
+if (currentLocale) {
+  locale = (adminLocale.language === currentLocale.language) ? adminLocale : currentLocale
+} else {
+  locale = adminLocale
+}
```
ariansobczak-rst, line 31: I think this would be simpler but I suggest storing locales in i18n resources. Regarding `useSelector` comment, this variable is outside store provider so the store is here relevant to  `window.REDUX_STATE` or `store.getState()`.

```suggestion
let locale = adminLocale
if (currentLocale) {
  locale = adminLocale.language === currentLocale.language ? adminLocale : currentLocale
}
```

src/frontend/bundle-entry.jsx
ariansobczak-rst: This variable is outside store provider so the store is here equivalent to `window.REDUX_STATE` or `store.getState()`

dziraf (approved):

PR #1457, opened 2023-04-03
fix(beta): use styled-components exported by @adminjs/design-system

dziraf:
BREAKING CHANGE: styled-components are now re-exported by @adminjs/design-system/styled-components to fix this library's issues with ESM


src/frontend/layout-template.tsx
```diff
@@ -51,6 +51,69 @@ const html = async (admin: AdminJS, currentAdmin?: CurrentAdmin, location = '/')
       <script src="${h.assetPath('app.bundle.js', assets)}"></script>
       <script src="${h.assetPath('components.bundle.js', assets)}"></script>
       ${styles.join('\n')}
+      <style>
```
ariansobczak-rst, line 54: Maybe we can put it into a separate css file like global.css in assets and link here 🤔 But I am wondering if it is necessary if it is a temporary resolution for the styled-components issue.

ariansobczak-rst (approved):

src/frontend/layout-template.tsx
dziraf: AdminJS's rollup is not configured to import css sadly. I would leave it as is until this issue is resolved

PR #1475, opened 2023-04-18
feat: store params in url with qs

ariansobczak-rst:
Replaced custom filter query params with https://www.npmjs.com/package/qs.

Allow developers to add queryParams in the action button like in exmaple below.
```tsx
<ActionButton
  action={newAction}
  resourceId={resource.id}
  queryParams={{
    email: 'example@example.com',
    isActive: true
  }}
>
  <Button variant="contained">
    <Icon icon="Plus" />
    {tb("addNewItem", resource.id)}
  </Button>
</ActionButton>;
```
Note: If passing queryParams to `new` action form will be preffield with url values i.e. using the example above form will have a filled email field with `example@example.com` and checked isActive.

Query params are stored in URL and can be accessible by using `useQueryParams`. There are separate params form lists like 
`sortBy`, `page`, `direction` or `filters` and also reference to `tab` or `redirectUrl`. To store params in url `storeParams` should be called with new params like:

```ts
const { storeParams } = useQueryParams()
...
storeParams({ page: 2 })
```

Parsed params from `useQueryParams` are all passed params parsed by `qs`.

For more information about `stringify` and `parse` functions please see qs configuration https://github.com/ljharb/qs.


src/frontend/components/actions/new.tsx
```diff
@@ -1,8 +1,12 @@
 import { Box, Button, DrawerContent, DrawerFooter, Icon } from '@adminjs/design-system'
+import identity from 'lodash/identity.js'
```
dziraf: several warnings such as:
```
'identity' is defined but never used
```

src/frontend/components/actions/new.tsx
```diff
@@ -15,25 +17,30 @@ import LayoutElementRenderer from './utils/layout-element-renderer.js'
...
     }
-  }, [initialRecord])
+  }, [initialRecord, parsedQuery])
 
-  const submit = (event: React.FormEvent<HTMLFormElement>): boolean => {
+  useEffect(() => {
+    if (parsedQuery) {
+      const resourceProperties = pick(parsedQuery, Object.keys(resource.properties))
+      if (Object.keys(resourceProperties).length) {
+        setRecord({ ...record, params: { ...record.params, ...resourceProperties } })
```
dziraf, line 35: What will happen if we add for example:

`?gender=abcd&isActive=false`?

If `gender` is `male`/`female`/`other` enum, will it set `abcd`?

If `isActive` is a boolean, will it correctly parse `true`/`false` strings?

src/frontend/components/app/filter-form.tsx
```diff
@@ -0,0 +1,90 @@
+import { Box, Button, DEFAULT_DRAWER_WIDTH, H3, Icon } from '@adminjs/design-system'
```
dziraf, line 1: `'Icon' is defined but never used`

src/frontend/components/routes/resource-action.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useState } from 'react'
 import { connect } from 'react-redux'
 import { useParams } from 'react-router'
```
dziraf, line 4: remove empty line or group Box with library imports

src/frontend/hooks/use-query-params.ts
```diff
@@ -0,0 +1,70 @@
+/* eslint-disable no-unused-vars */
+import isEmpty from 'lodash/isEmpty.js'
```
dziraf, line 2: let's add `src/frontend/hooks/index.js` and export all hooks from it

src/frontend/components/app/filter-form.tsx
ariansobczak-rst: I removed this component because it was only for local development and maybe can be introduced in the future to allow users to show filters not as a drawer but more as a side panel of list

image

src/frontend/components/actions/new.tsx
ariansobczak-rst: Done

src/frontend/components/routes/resource-action.tsx
ariansobczak-rst: Done

src/frontend/hooks/use-query-params.ts
ariansobczak-rst: Done

src/frontend/components/actions/new.tsx
ariansobczak-rst: Checked with enums and booleans and there are results

**for enums:**
value in url has to be exact like availableValues e.g.
```
export enum CountryEnum {
  Poland = 'PL',
  Germany = 'DE',
}
```
should look like `?country=PL` (case sensitive). Otherwise value would not be filled in new item form. 
image
image

**for booleans**
Booleans are correctly parsed by qs parse
image

PR #1485, opened 2023-04-27
feat: allow to clear query params by keys

ariansobczak-rst:


src/frontend/hooks/use-query-params.ts
```diff
@@ -50,11 +54,16 @@ export function useQueryParams<FiltersT = Record<string, unknown>>() {
...
   )
 
-  const storeParams = useCallback((params: Partial<Params>) => {
+  const storeParams = useCallback((params: Partial<Params<ParamsT, FiltersT>>) => {
     const newQuery = { sortBy, direction, page, tab, filters, redirectUrl, ...params }
     return setSearchParams(stringify(newQuery, { skipNulls: true, allowDots: true }))
   }, [])
 
+  const clearParams = useCallback((...params: (keyof Params<ParamsT, FiltersT>)[]) => {
+    const newQuery = params.reduce((memo, curr) => ({ ...memo, [curr]: undefined }), {})
```
dziraf, line 63: this won't result in `refresh=undefined` right?

dziraf (approved):

src/frontend/hooks/use-query-params.ts
ariansobczak-rst: Replaced with `searchParams.delete` to be sure that param will be removed from URL instead of assigning undefined.

dziraf (approved):

PR #1501, opened 2023-05-11
fix: fix back button not closing drawer

jwasiak:


src/frontend/components/app/drawer-portal.tsx
```diff
@@ -74,11 +74,11 @@ const getOrCreatePortalContainer = (id: string) => {
  */
 export const DrawerPortal: React.FC<DrawerPortalProps> = ({ children, width }) => {
   const [drawerElement, setDrawerElement] = useState(document.getElementById(DRAWER_PORTAL_ID))
-  const { to = null } = useSelector<ReduxState, RouterInState>((state) => state.router)
+  const { from = null } = useSelector<ReduxState, RouterInState>((state) => state.router)
```
dziraf, line 77: Are you sure `from` will work? If I remember correctly, there was some kind of double navigation and `from` wasn't the latest route here 🤔

dziraf (approved):

PR #1527, opened 2023-07-10
Updated notice to be able to close from Phones

ReisoAphi:
Updated the width to make it be able to close the notice from small phones.


dziraf:
Wouldn't it make the notice box use the entire screen width on desktops though?

dziraf:
Closing due to lack of response

PR #1544, opened 2023-08-15
fix: Fixes the back button route in drawer

vinu:


dziraf:
@vinu I suggested a different fix here: https://github.com/SoftwareBrothers/adminjs/pull/1545

Changing `from` to `to` wouldn't help because it in turn breaks bulk actions that also open a drawer (it used to be `to` before).

vinu:
@dziraf - I am closing this :) thanks for the fix

Issue #1543, opened 2023-08-15
[Bug]: Backbutton in the drawer when action.showInDrawer = true routes to invalid route


vinu:
### What happened?

When you have resources with 
``` javascript
actions: {
  show: {
    showInDrawer: true,
},
```
The show opens in the drawer, but the back button goes to an invalid route. 

### Bug prevalence

When ever the show action is opened in a drawer

### AdminJS dependencies version

"@adminjs/express": "^6.0.0",
"@adminjs/mongoose": "^4.0.0",
"adminjs": "^7.2.0",
"express": "^4.18.2",
"express-formidable": "^1.2.0",
"express-session": "^1.17.3",
"mongoose": "7.0.1",
"nodemon": "^3.0.1",
"tslib": "^2.6.1"

PR #1545, opened 2023-08-16
fix: make back button navigate one route up

dziraf:


ariansobczak-rst (approved):

dziraf:
This fixes the issue where drawer doesn't close correctly, but introduces another issue where refreshing the page will result in 404. This is because `to=".." relative="path"` only removes URL's pathname up to last `/` which in case of `show` action removes record's ID which results in drawer getting closed. AdminJS doesn't have a record route without `:recordId` param defined though, so when you refresh the page you see a `CANNOT GET ...` error.

We'd probably have to rewrite routing so that routes are nested, this way we could simply use `relative="route"`.

vinu:
Hmm but will this work? with filters? 

I just tried like this

```jsx
<StyledLink
      size="icon"
      onClick={() => navigate(-1)}
      rounded
      mr="lg"
      type="button"
    >
```
Seems working

vinu:
nah - disregard that - it won't work with multiple show click, - the back click just loads the previous show action

dziraf:
I rewrote routing so that back button works as expected now, but it turned out we also have to rewrite the Drawer component since it doesn't show list underneath anymore.


https://github.com/SoftwareBrothers/adminjs/assets/16668924/d8e1b0cd-a044-4934-8c6a-c771c9de44f5

dziraf:
It looks to be working correctly now

https://github.com/SoftwareBrothers/adminjs/assets/16668924/63b9ceda-58d3-41b4-9d36-4ff67c5d1024

src/frontend/interfaces/action/build-action-click-handler.ts
```diff
@@ -65,7 +67,17 @@ export const buildActionClickHandler = (
 
     // Default behaviour - you're navigated to action URL and logic is performed on it's route
     if (href) {
-      navigate(href)
+      const url = new URL(`relative:${href}`)
```
dziraf, line 70: appending pathname with `<string>:` is a hack which allows us to use `new URL(...)` if we only have a relative url

ariansobczak-rst: 👍🏻 Nice, I didn't know that.

ariansobczak-rst (approved):

dziraf:
fixes https://github.com/SoftwareBrothers/adminjs/issues/1543

Issue #1473, opened 2023-04-18
AdminJS v7 - feedback, urgent issues, questions, discussions


dziraf:
This is a thread to aggregate all questions/feedback/urgent bugs/discussions related to the recently released version 7.

Feel free to share your ideas or let us know about any issues you're facing related to migration or usage.

**[What's new in version 7
](https://docs.adminjs.co/installation/whats-new-in-v7)**

**[Migration Guide](https://docs.adminjs.co/installation/migration-guide-v7)**

If you notice that parts of our documentation are outdated or unclear, please share that in this thread as well.

ddresch:
Working on a migration of a super basic app with just 4-5 resources and mongoose schema setup. Did the pnpm / npm / yarn update and now the system can't find the packages anymore. 

https://www.npmjs.com/package/@adminjs/mongoose the docs are wrong - still showing require statements.

```js
import { Database as MongooseDatabase, Resource as MongooseResource } from '@adminjs/mongoose';
```

Using this like in the new example app is dropping the error I'm mentioning above.

mszula:
Great work guys!

I'm really in favor of moving to ESM-only but I feel like one thing hasn't been thought through. It's cooperation with NestJS which still uses CJS and has no plans to move to ESM.

https://github.com/nestjs/nest/issues/7021

Do you have any solution for that?

dziraf:
@mszula until Nest.js moves to ESM or starts dual packaging, you will either have to set up AdminJS express/fastify way:

**src/app.module.ts**
```typescript
import { Module } from '@nestjs/common';

import { AppController } from './app.controller';
import { AppService } from './app.service';

@Module({
  imports: [
    import('@adminjs/nestjs').then(({ AdminModule }) =>
      AdminModule.createAdminAsync({
        useFactory: () => ({
          adminJsOptions: { rootPath: '/' },
        }),
      }),
    ),
  ],
  controllers: [AppController],
  providers: [AppService],
})
export class AppModule {}
```

**tsconfig.json**
```json
{
  "compilerOptions": {
    // ...
    "moduleResolution": "nodenext"
  }
}
```

Instead of `import(...).then(...)` in `app.module.ts` you can just set it up Express way in **main.ts**:
```typescript
import { NestFactory, HttpAdapterHost } from '@nestjs/core';

import { AppModule } from './app.module';

async function bootstrap() {
  const app = await NestFactory.create(AppModule);

  const AdminJS = await import('adminjs');
[...]

PR #1596, opened 2023-12-22
fix: FilterDrawer UI issue

AstaDK:
This change to fix the issue filter drawer UI
Detail issue here: https://github.com/SoftwareBrothers/adminjs/issues/1473#issuecomment-1865387328


dziraf:
I think the filter UI is now closed by default

PR #1600, opened 2024-01-02
fix: return existing bundle immediately

mohd-akram:
Skip generating the entry file and calling mkdir if not needed.


mohd-akram:
Thoughts on merging this? Currently the function might modify the filesystem even when `write` is false, which I don't think is desirable. This also avoids generating the entry file every time despite not being needed, speeding up the request.

PR #1604, opened 2024-01-15
fix: list action populate issue

AstaDK:
Explain issue:

I'm facing an issue with a resource list. The error message reads: "TypeError: Cannot read properties of undefined (reading 'find')." I have attached an image for more details. 

I suspect that the issue is due to a removed relationship.

"@adminjs/express": "^6.0.1",
"@adminjs/mongoose": "^4.0.0",
"adminjs": "^7.5.2",

image


dziraf (approved):

dziraf:
Hi @AstaDK, I'll approve and merge the PR since it should fix the UI error. However, I think you might have a configuration issue there since `populated.recordActions` should actually be defined and default to an empty array if the reference is configured correctly.

PR #1607, opened 2024-01-19
fix: add missing translations for boolean/select fields

dziraf:


src/backend/decorators/property/property-options.interface.ts
```diff
@@ -66,7 +66,7 @@ export default interface PropertyOptions {
    */
   availableValues?: Array<{
     value: string | number;
-    label: string;
+    label?: string;
```
dziraf, line 69: We should remove `label` in v8 since all translations are done in the frontend now and keeping a label here seems to be pointless.

PR #1628, opened 2024-02-13
fix: fix records list not visible when resource action is open in drawer

dziraf:


src/frontend/components/routes/resource-action.tsx
```diff
@@ -37,13 +38,44 @@ const ResourceAction: React.FC<Props> = (props) => {
     return <NoActionError resourceId={resourceId!} actionName={actionName!} />
   }
 
+  const listActionName = 'list'
+  const listAction = resource.resourceActions.find((r) => r.name === listActionName)
+
   const contentTag = getResourceElementCss(resource.id, action.name)
 
   if (action.showInDrawer) {
+    console.log('list action', listAction)
```
ariansobczak-rst, line 47: Is this log intentional?

dziraf: nope, removed

PR #1631, opened 2024-02-20
feat: update rollup to v4, use esbuild to minify instead of terser

dziraf:


.github/workflows/push.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - name: Setup
         uses: actions/setup-node@v3
         with:
-          node-version: '18'
+          node-version: '18.17'
```
dziraf, line 14: Mocha fails with random error on higher node versions

src/adminjs.ts
```diff
@@ -39,12 +45,12 @@ export const defaultOptions: AdminJSOptionsWithDefault = {
 }
 
 type ActionsMap = {
-  show: Action<RecordActionResponse>;
-  edit: Action<RecordActionResponse>;
-  delete: Action<RecordActionResponse>;
-  bulkDelete: Action<BulkActionResponse>;
-  new: Action<RecordActionResponse>;
-  list: Action<ListActionResponse>;
+  show: Action<RecordActionResponse>
```
ariansobczak-rst, line 48: ❓ Is this some new Eslint rule to remove semicolons?

ariansobczak-rst (approved):

src/adminjs.ts
dziraf: IDK, I have autoformatting enabled on save and it does that

PR #1643, opened 2024-03-28
feat: support subproperties as a title

smacker:
This change allows setting a sub property as a title of resource.

For example:
```ts
      resource: {
        model: getModelByName('Sport'),
        client: prisma,
      },
      options: {
        titleProperty: 'translation.en',
        listProperties: ['translation.ja', 'translation.en'],
        properties: {
          translation: { type: 'mixed' },
          'translation.ja': { type: 'string' },
          'translation.en': { type: 'string' },
          // some other props
        },
      },
```

In master branch such `titleProperty` is just ignored which is an unexpected behavior.

### DIFF
--- a/src/backend/decorators/resource/resource-decorator.ts
+++ b/src/backend/decorators/resource/resource-decorator.ts
@@ -250,7 +250,7 @@ class ResourceDecorator {
 
     const properties = Object.values(this.properties)
     if (this.options.titleProperty) {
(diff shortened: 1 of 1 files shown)


dziraf (approved):

dziraf:
@smacker Thanks, I'll merge this. Using the helper method actually makes more sense here.

PR #1665, opened 2024-05-19
fix: running with Node.js 22

### LINKED ISSUE #1663: [Bug]: Use import specifier 'with' instead of 'assert'
### Contact Details

_No response_

### What happened?

Since 'assert' has been removed from the spec in favor of 'with' ( see https://nodejs.org/api/esm.html#import-specifiers ) building on NodeJS 22 will throw the following error:

SyntaxError: Unexpected identifier 'assert'

image

All you have to do to fix is change 'assert' to 'with'

image

### Bug prevalence

Always on NodeJS 22

### AdminJS dependencies version

"adminjs": "^7.8.1",
    "@adminjs/bundler": "^3.0.0",
    "@adminjs/design-system": "^4.1.1",
    "@adminjs/express": "^6.1.0",
    "@adminjs/import-export": "^3.0.0",
    "@adminjs/themes": "^1.0.1",
    "@adminjs/typeorm": "^5.0.1",

### What browsers do you see the problem on?

_No response_

### Relevant log output

_No response_

### Relevant code that's giving you issues

_No response_

mohd-akram:
Resolves `SyntaxError: Unexpected identifier 'assert'` in Node.js 22.

Closes #1663

### DIFF
--- a/.babelrc.json
+++ b/.babelrc.json
@@ -19,5 +19,8 @@
     "src/frontend/assets/scripts/app-bundle.production.js",
     "src/frontend/assets/scripts/global-bundle.development.js",
     "src/frontend/assets/scripts/global-bundle.production.js"
-  ]
+  ],
+  "generatorOpts": {
+    "importAttributesKeyword": "with"
+  }
 }
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Setup
         uses: actions/setup-node@v3
         with:
-          node-version: '18.17'
+          node-version: '18'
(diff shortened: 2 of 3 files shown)


dziraf:
Changing the import syntax is not enough, there are extra changes to Babel required + the last time I checked TS highlighted `with` as an error even though we're already at v5.3+ so it'd be the best to update `typescript` too.

mohd-akram:
Support for import attributes was added in Node.js [18.20](https://nodejs.org/en/blog/release/v18.20.0) so the CI version needed to be updated.

dziraf (approved):

PR #1694, opened 2024-07-31
fix: propertyType date edit should not convert to local timezone

AshotN:
Similar to #1692 but for the edit component. 

Should have no impact on `date-time` component as that was already working correctly.

### DIFF
--- a/src/frontend/components/property-type/datetime/edit.tsx
+++ b/src/frontend/components/property-type/datetime/edit.tsx
@@ -23,7 +23,7 @@ const Edit: React.FC<EditPropertyProps> = (props) => {
         value={value}
         disabled={property.isDisabled}
         onChange={(date) => {
-          onChange(property.path, date?.substring(0, 10) ?? date)
+          onChange(property.path, property.type === 'date' ? date?.substring(0, 10) ?? date : date)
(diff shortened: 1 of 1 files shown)


src/frontend/components/property-type/datetime/edit.tsx
```diff
@@ -6,10 +6,13 @@ import { recordPropertyIsEqual } from '../record-property-is-equal.js'
 import { PropertyLabel } from '../utils/property-label/index.js'
 import allowOverride from '../../../hoc/allow-override.js'
 import { useTranslation } from '../../../hooks/index.js'
+import { PropertyType } from '../../../../backend/index.js'
+
+const formatDate = (val:any, propertyType: PropertyType) => (propertyType === 'date' ? `${val}T00:00:00` : val)
```
dziraf, line 11: `val` should probably be just a `string | null` since the data comes from the api. We probably have to check for `null` to so it doesn't convert to `nullT00:00:00`

dziraf (changes requested):

AshotN:
@dziraf anything else holding this PR?

dziraf (approved):

Issue #1691, opened 2024-07-31
[Bug]: date with no time format should not convert to local timezone


AshotN:
### What happened?

The issue is when an adapter passes a value such as '2024-07-31' to AdminJS, the frontend will display it in local time(`2024-07-30`). Causing the UI to not match what the database is actually storing.

I propose fixing this by modifying this file (https://github.com/SoftwareBrothers/adminjs/blob/master/src/frontend/components/property-type/datetime/map-value.ts#L9)

Adding a check like this would resolve the problem

```ts
const date = propertyType === 'date' ? new Date(`${value}T00:00:00`) : new Date(value)
```

Semi-Related: https://github.com/SoftwareBrothers/adminjs/issues/978

### Bug prevalence

Always

### AdminJS dependencies version

```
"adminjs": "7.8.7",
```

PR #1700, opened 2024-08-13
fix: fix property type 'date' when backend returns an iso string

dziraf:
Fixes issue related to https://github.com/SoftwareBrothers/adminjs/issues/1691

@AshotN this fixes a bug where date values were rendered as `NaN-NaN-NaN` if backend returned an ISO string instead of a simple date string while having the property's `type` set to `date` instead of `datetime`.

### DIFF
--- a/src/frontend/components/property-type/datetime/edit.tsx
+++ b/src/frontend/components/property-type/datetime/edit.tsx
@@ -7,10 +7,14 @@ import { PropertyLabel } from '../utils/property-label/index.js'
 import allowOverride from '../../../hoc/allow-override.js'
 import { useTranslation } from '../../../hooks/index.js'
 import { PropertyType } from '../../../../backend/index.js'
+import { stripTimeFromISO } from './strip-time-from-iso.js'
 
-const formatDate = (val:string|null, propertyType: PropertyType) => {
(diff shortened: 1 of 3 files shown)


AshotN:
Makes sense, I haven't been able to test on my actual project yet. But looking at the changes it looks good to me. Sorry for introducing that issue

dziraf:
I'll have to give this more thought. If the database/backend stores datetimes using UTC timezone and the user's timezone is i.e. GMT+4, this will convert backend's 2024/08/13 11:00PM to 2024-08-13, while that should be 2024-08-14 in the user's timezone.

AshotN:
My thought process was for date's with no time. We should display exactly what the DB is storing, so `2024-01-01` no matter where you are should display `2024-01-01`. For editing I can see it either way, if your local timezone should be converted to UTC, or just write it as you input it. The latter is more intuitive at face value. But I'm not sure what people would expect more

phprelated:
Hi there,
Upgrading from 7.8.1 to 7.8.12 gave me the NaN-NaN-NaN issue.
Waiting for this fix to be deployed.
Thanks

dziraf:
I'll merge this for now. We should consider using https://www.npmjs.com/package/@formkit/tempo for dates in the meantime

PR #1721, opened 2024-12-06
feat(new): Add ability to set the newProperties

ChristopheCVB:
Add ability to set the `newProperties` the same way we set the `editProperties`

### DIFF
--- a/src/backend/decorators/resource/resource-decorator.ts
+++ b/src/backend/decorators/resource/resource-decorator.ts
@@ -318,6 +318,9 @@ class ResourceDecorator {
       editProperties: this.getProperties({
         where: 'edit',
       }).map((property) => property.toJSON('edit')),
+      newProperties: this.getProperties({
+        where: 'new',
+      }).map((property) => property.toJSON('new')),
       showProperties: this.getProperties({
         where: 'show',
       }).map((property) => property.toJSON('show')),
--- a/src/frontend/components/actions/new.tsx
+++ b/src/frontend/components/actions/new.tsx
@@ -87,7 +87,7 @@ const New: FC<ActionProps> = (props) => {
               record={record as RecordJSON}
             />
           ))
-          : resource.editProperties.map((property) => (
+          : resource.newProperties.map((property) => (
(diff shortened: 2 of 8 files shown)


src/backend/decorators/resource/resource-decorator.ts
```diff
@@ -318,6 +318,9 @@ class ResourceDecorator {
       editProperties: this.getProperties({
         where: 'edit',
       }).map((property) => property.toJSON('edit')),
+      newProperties: this.getProperties({
+        where: 'new',
```
dziraf, line 322: You might have to extend the type here:
https://github.com/SoftwareBrothers/adminjs/blob/master/src/frontend/interfaces/property-json/property-json.interface.ts#L3

Thanks for contributing; I'll try to take a proper look at this PR next week.

src/frontend/components/actions/new.tsx
```diff
@@ -87,7 +87,7 @@ const New: FC<ActionProps> = (props) => {
               record={record as RecordJSON}
             />
           ))
-          : resource.editProperties.map((property) => (
+          : resource.newProperties.map((property) => (
```
dziraf, line 90: We might also need to update `where` to `new` in `BasePropertyComponent` and `LayoutElementRenderer`. `new` would then also have to be added to i. e. `isVisible` for properties etc.

src/backend/decorators/resource/resource-decorator.ts
ChristopheCVB: done ;)

src/frontend/components/actions/new.tsx
ChristopheCVB: done too

ChristopheCVB:
@dziraf Any chance you have a look at this soon?

src/frontend/components/actions/new.tsx
ChristopheCVB: The where is used to determine the component to render in `BasePropertyComponent` and we actually want to render the property with the edit component

PR #1735, opened 2025-01-21
add div closing tag

Mahad-lab:


Mahad-lab (commented):
add closing tag to div to fix html error

aszymanskiit (commented):
Self-closing tags are completely valid in TSX, and they follow the expected JSX syntax. Since they are required for elements without children, there's no issue with using them.

PR #1750, opened 2025-03-05
feat(filter): add support for more operators than contains

roggervalf:
When filtering data, by default, default-type component generates urls where records are going to be filtered by attribute using contains operations. This pr provides a way to choose different operators and reflect that in the url. Operators are: startsWith, endsWith, equals, notEquals and contains (default) operators. Or operator is supported too.
Screenshot 2025-03-17 at 1 15 41 a m
Updates were added to these adapters as well:
* [Sequelize](https://github.com/SoftwareBrothers/adminjs-sequelizejs/pull/126)
* [Objectionjs](https://github.com/SoftwareBrothers/adminjs-objection/pull/11)
* [Mongoose](https://github.com/SoftwareBrothers/adminjs-mongoose/pull/117)
* [Typeorm](https://github.com/SoftwareBrothers/adminjs-typeorm/pull/76)
* [Prisma](https://github.com/SoftwareBrothers/adminjs-prisma/pull/60)
* [Mikroorm](https://github.com/SoftwareBrothers/adminjs-mikroorm/pull/24)

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
   "homepage": "https://github.com/SoftwareBrothers/adminjs#readme",
   "dependencies": {
     "@adminjs/design-system": "^4.1.0",
+    "@adminjs/themes": "^1.0.1",
     "@babel/core": "^7.23.9",
     "@babel/parser": "^7.23.9",
     "@babel/plugin-syntax-import-assertions": "^7.23.3",
--- a/package.json
+++ b/package.json
@@ -86,7 +86,6 @@
   "homepage": "https://github.com/SoftwareBrothers/adminjs#readme",
   "dependencies": {
     "@adminjs/design-system": "^4.1.0",
-    "@adminjs/themes": "^1.0.1",
     "@babel/core": "^7.23.9",
     "@babel/parser": "^7.23.9",
     "@babel/plugin-syntax-import-assertions": "^7.23.3",
--- a/src/frontend/components/property-type/default-type/filter.tsx
+++ b/src/frontend/components/property-type/default-type/filter.tsx
@@ -1,17 +1,26 @@
-import React from 'react'
-import { FormGroup, Input, Select } from '@adminjs/design-system'
+import React, { useState } from 'react'
+import { Box, FormGroup, Input, Select } from '@adminjs/design-system'
 
 import allowOverride from '../../../hoc/allow-override.js'
 import { FilterPropertyProps } from '../base-property-props.js'
 import PropertyLabel from '../utils/property-label/property-label.js'
 import { useTranslation } from '../../../hooks/use-translation.js'
+import * as BackendFilter from '../../../../backend/utils/filter/filter.js'
+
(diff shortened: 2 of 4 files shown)


roggervalf:
pls @dziraf when you get some time to take a look

package.json
```diff
@@ -86,6 +86,7 @@
   "homepage": "https://github.com/SoftwareBrothers/adminjs#readme",
   "dependencies": {
     "@adminjs/design-system": "^4.1.0",
+    "@adminjs/themes": "^1.0.1",
```
roggervalf, line 89: added as when using example repo https://github.com/SoftwareBrothers/adminjs-example-app fails when this package is linked and this package failed because this dependency does not exist

src/frontend/components/property-type/default-type/filter.tsx
```diff
@@ -1,27 +1,65 @@
...
 import { useTranslation } from '../../../hooks/use-translation.js'
+import * as BackendFilter from '../../../../backend/utils/filter/filter.js'
+
+const { PARAM_SEPARATOR } = BackendFilter
 
 const Filter: React.FC<FilterPropertyProps> = (props) => {
   const { property, onChange, filter } = props
+
+  const possibleKeys = [
+    property.path,
```
roggervalf, line 16: keeping contains key as it was before to prevent other adapters to not be able to recognize it

package.json
dziraf: This will result in a circular dependency issue because `adminjs` is a peer dependency of `@adminjs/themes`. `adminjs-example-app` is also an outdated repository.

Local linking issue can be solved by linking `@adminjs/themes` in your local app and in local `adminjs`

roggervalf:
Hi @dziraf, thanks for takint a look. I like the idea. Wonder if I can open separate prs for the other operators after this one to support them pr do you prefer all changes at once?

package.json
roggervalf: good to know, I'll remove this package though

roggervalf:
> This is a good start but ideally the entire filter system should be rebuilt to allow `AND`, `OR` and `NOT` plus we should use symbols for operators. This would also require us to update all adapters to make sure the app works the same regardless of your ORM/ODM/database

hey @dziraf, I added this functionality into adapaters and covered the other operators that you mentioned. Pls when you get some time

roggervalf:
> This is a good start but ideally the entire filter system should be rebuilt to allow `AND`, `OR` and `NOT` plus we should use symbols for operators. This would also require us to update all adapters to make sure the app works the same regardless of your ORM/ODM/database

hey @dziraf, I added this functionality into adapters and covered the other operators that you mentioned. Pls when you get some time

Visot:
Having support for multiple operators would be really useful. It adds flexibility to how we filter data and makes the component much more versatile in different use cases.

roggervalf:
hey @dziraf, could you pls help me with your Review when you get some time

diyorbekrustamjonov (approved):

PR #1787, opened 2025-09-25
fix for non-mixed array properties

serjou1:


serjou1:
@dziraf please check and approve

PR #1796, opened 2025-11-16
fix: correct frontend asset paths

badzim:
This PR fixes a path-resolution issue in the AdminJS router that causes all frontend bundles (`app.bundle.js`, `global.bundle.js`, `design-system.bundle.js`) to return **404** when AdminJS is installed with **pnpm**, especially in monorepos or virtual store setups.

Because the current code relies on `__dirname` + relative segments to locate `lib/frontend/assets`, the computed paths point to the wrong location under pnpm. As a result, the AdminJS UI does not load — this also affects the **AdminJS demo site**, which is currently down for the same reason.

This PR replaces the brittle relative paths with robust module-resolution logic:

```ts
const ASSETS_ROOT = path.join(
  process.cwd(),
  'node_modules',
  'adminjs',
  'lib',
  'frontend',
  'assets',
)

const resolveDesignSystemBundle = () => {
  return path.join(
    process.cwd(),
    'node_modules',
    '@adminjs',
    'design-system',
    `bundle.${NODE_ENV}.js`,
  )
}
```

This approach works reliably across:

* pnpm (hoisted and non-hoisted)
* npm / yarn
* monorepos / workspaces
* Vite / NestJS / Express integrations

With this fix, all asset URLs under `/admin/frontend/assets/*` resolve correctly again, and the AdminJS UI loads normally.


badzim:
@dziraf what do you think ?

